PR #1, opened 2020-04-28
Adds send-dependency-update workflow

ryanmoran:
This workflow is the first half of the process to get implementation
updates into language family buildpacks. It consists of 3 parts:

1. An action called "dependency/parse" that reads in a release published
   event payload and uses that to fetch all of the details required to
   update a dependency (ie. id, version, stacks, source/uri, and sha
   checksums). These details are returned as outputs of the action.

2. An action called "dispatch" that sends a repository_dispatch event to
   a repository. This action is generic and can be used for all sorts of
   mechanisms to link together sets of actions across repos.

3. A workflow for implementation buildpacks called
   "send-dependency-update" that links these actions together to perform
   the process to notifying a language-family buildpack when the
   implementation buildpack releases a new version.

Relates to https://github.com/paketo-buildpacks/nodejs/issues/1


implementation/workflows/send-dependency-update.yml
```diff
@@ -0,0 +1,44 @@
+# This workflow triggers on a published release event.
+# Requires the following secrets:
+# - LANGUAGE_FAMILY_REPO -> the full name of the repo to send the dispatch to (eg. paketo-buildpacks/nodejs)
```
arjun024, line 3: Potentially this could be a list and the dispatch action calls multiple dispatchers

PR #4, opened 2020-05-05
Use mergify app to auto-merge PRs in lang family

thitch97:
- These are the PRs created by github actions when
new implementation buildpacks are released

Co-authored-by: Arjun Sreedharan <asreedharan@vmware.com>

Relates to https://github.com/paketo-buildpacks/nodejs/issues/1

### DIFF
--- a/language-family/.mergify/config.yml
+++ b/language-family/.mergify/config.yml
@@ -8,4 +8,4 @@ pull_request_rules:
       merge:
         strict: true
         method: rebase
-      delete_head_branch:
+      delete_head_branch: {}
--- a/language-family/.mergify/config.yml
+++ b/language-family/.mergify/config.yml
@@ -1,0 +1,11 @@
+pull_request_rules:
+  - name: Auto merge pull requests from github-actions bot
+    conditions:
+      - author=github-actions[bot]
+      # TBD when language-family has integration/acceptance tests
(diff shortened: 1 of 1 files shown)


language-family/.mergify/config.yml
```diff
@@ -0,0 +1,11 @@
...
+  - name: Auto merge pull requests from github-actions bot
+    conditions:
+      - author=github-actions[bot]
+      # TBD when language-family has integration/acceptance tests
+      # - status-success=Integration Test
+    actions:
+      merge:
+        strict: true
+        method: rebase
+      delete_head_branch:
```
ryanmoran, line 11: Is this field intentionally set to `null`?

arjun024:
(Co-author)
This PR's file structure assumes that at the end of the day, the contents of directories `implementation` and `language-family` would be a structured as a subset of the **root** of their target repositories. Like as follows:

```bash
├── implementation
│   ├── .github
│       ├── util
│       ├── workflows
└── language-family
    ├── .github
    │   ├── util
    │   ├── workflows
    ├── .mergify
        ├── config.yml
```

language-family/.mergify/config.yml
arjun024: Yes. Can change to `delete_head_branch: {}` if that's more readable.

PR #7, opened 2020-05-06
Fixes typo in error message

ryanmoran:


arjun024:
We have removed the capitalization of "error" as part of #9 if that's good enough

ryanmoran:
@arjun024 its not about the capitalization. its about that it would appear as "Error: error: something" because of how `fail` is implemented: https://github.com/paketo-buildpacks/github-config/blob/ea2e7272e1108582469e6aac20ccfe2bfbe60f39/actions/dispatch/entrypoint/main.go#L83-L86

PR #10, opened 2020-05-06
language family package.sh script

dwillist:
- creates a .cnb artifact from a buildpacks buildpack.toml file.

### DIFF
--- a/language-family/scripts/package.sh
+++ b/language-family/scripts/package.sh
@@ -37,7 +37,6 @@ function main {
   os::check
   repo::prepare
 
-  github::token::fetch
   tools::jam::install
   tools::yj::install
   tools::pack::install
@@ -79,7 +78,7 @@ function repo::prepare() {
 }
 
 function tools::jam::install() {
-  echo "-> Installing latest jam..."
+  echo "-> Installing v0.0.7 jam..."
 
   local os
   os="${OS}"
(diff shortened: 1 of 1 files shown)


language-family/scripts/package.sh
```diff
@@ -0,0 +1,191 @@
...
+      --package-config "${BUILD_DIR}/package.toml" \
+      --format file
+}
+
+function github::token::fetch() {
+  if [[ -z "${GITHUB_TOKEN:-""}" ]]; then
+    echo "-> Fetching GITHUB_TOKEN..."
+
+    GITHUB_TOKEN="$(
+      lpass show Shared-CF\ Buildpacks/concourse-private.yml \
```
arjun024, line 182: Would it be better to hard-code versions of jam, pack etc now so that you don't need lpass private bucket access, and what's tested in a PR will be the same as the one in the base branch?

language-family/scripts/package.sh
```diff
@@ -0,0 +1,191 @@
...
+}
+
+function buildpackage::toml::write() {
+  echo "-> Generating package config in ${BUILD_DIR}/package.toml..."
+
+  yj -tj < "${ROOT_DIR}/buildpack.toml" \
+    | jq -r '.metadata.dependencies[] | select(.id != "lifecycle") | {uri: .uri }' \
+    | jq -s --arg uri "${BUILD_DIR}/buildpack.tgz" '. | {buildpack: {uri: $uri}, dependencies: .}' \
+    | yj -jt \
+    > "${BUILD_DIR}/package.toml"
```
arjun024, line 165: Is there a place that defines what package.toml is? Couldn't find in any of the specs or pack docs

ForestEckhardt: @arjun024 https://buildpacks.io/docs/reference/package-config/

arjun024: Thanks @ForestEckhardt

language-family/scripts/package.sh
ryanmoran: That seems reasonable. We will change this.

PR #19, opened 2020-05-11
Changes versioning logic

ForestEckhardt:
This change make it so that the version sting is v prepended for the github tag and release name but no where else. This was causing issues creating buildpackages because the versions that we provided in buildpack.tomls of the implementation buildpacks and meta buildpacks did not line up.

I decided to keep the v prepended to the tag and title on github on after reading [this](https://semver.org/#is-v123-a-semantic-version) from the semver spec.

### DIFF
--- a/actions/tag/entrypoint
+++ b/actions/tag/entrypoint
@@ -6,9 +6,8 @@
 main() {
   previous="$(git describe --tags "$(git rev-list --tags --max-count=1)")"
   tag="$(printf "%s" "$previous" | awk -F. '{$NF = $NF + 1;} 1' | sed 's/ /./g')"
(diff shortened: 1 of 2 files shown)


arjun024:
@ForestEckhardt. And the `buildpack.toml` will pick up the version from the git tag. Is that right?

ForestEckhardt:
@arjun024  The `v` will only ever appear in the github release. The smever standard indicates that  a prepended `v` is not valid semver, but the `v` prepend is allowable for version control systems. I  think that mixing the two a `v` prepend on github (which I think gives improved readability) and no `v` in `buildpack.toml` (which should be a semver valid string) gives us what we want.

PR #25, opened 2020-05-19
Adds detection for single binary builds in build.sh

ForestEckhardt:


implementation/scripts/build.sh
```diff
@@ -6,20 +6,44 @@ readonly PROGDIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
...
+
+            for name in detect build; do
+              printf "%s" "Linking ${name}..."
+
+              ln -sf "run" "${name}"
+
+              echo "Success!"
+            done
+        popd > /dev/null || return
+    else
```
ryanmoran, line 31: I don't think that these cases are mutually exclusive. For example, once we move `php-web` over to `packit` with a single binary structure, there will still be 2 extra commands in the `cmd` directory that we would want to build as can be seen here: https://github.com/paketo-buildpacks/php-web/tree/master/cmd.

I think you want to do the first branch iff there is a `main.go`, but then always to the second branch because some future CNBs might have both forms.

ForestEckhardt: Ok sounds good, I will make that change asap.

ForestEckhardt: @ryanmoran I need to still get the build on the `cmd` directory existing otherwise I get an error but I am pretty sure that this does everything that you want.

PR #83, opened 2020-06-29
allow the sync action to work with bootstrapper

joshzarrabi:
Signed-off-by: Daniel Thornton <dthornton@vmware.com>

### DIFF
--- a/actions/sync/entrypoint
+++ b/actions/sync/entrypoint
@@ -3,6 +3,7 @@ set -euo pipefail
 
 config_repo=
 current_repo=
+current_path=
 config_path=
 branch_name=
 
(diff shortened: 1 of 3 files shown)


actions/sync/entrypoint
```diff
@@ -37,11 +38,12 @@ sync_dirs() {
   set -x
   cd "${config_repo}${config_path}"
     for dir in $(ls -Ap . | grep "/"); do
-      if [[ -f "${current_repo}/${dir}.syncignore" ]]; then
-        rsync -r "${dir}" "${current_repo}/${dir}" --delete \
-          --exclude=".syncignore" --exclude-from="${current_repo}/${dir}.syncignore"
+      local exclude_list=""
+      if [[ -f "${current_repo}${current_path}/${dir}.syncignore" ]]; then
+        rsync -r "${dir}" "${current_repo}${current_path}/${dir}" --delete \
+          --exclude=".syncignore" --exclude-from="${current_repo}${current_path}/${dir}.syncignore"
```
ryanmoran, line 44: You should always include a `/` between path segments. Extra `/` characters back-to-back do not hurt, but missing `/` characters means that your script breaks.

```suggestion
      if [[ -f "${current_repo}/${current_path}/${dir}.syncignore" ]]; then
        rsync -r "${dir}" "${current_repo}/${current_path}/${dir}" --delete \
          --exclude=".syncignore" --exclude-from="${current_repo}/${current_path}/${dir}.syncignore"
```

actions/sync/entrypoint
```diff
@@ -37,11 +38,12 @@ sync_dirs() {
...
       else
-        rsync -r "${dir}" "${current_repo}/${dir}" --delete
+        rsync -r "${dir}" "${current_repo}${current_path}/${dir}" --delete
```
ryanmoran, line 46:
```suggestion
        rsync -r "${dir}" "${current_repo}/${current_path}/${dir}" --delete
```

joshzarrabi: I think we don't want it just incase its an empty string?

ryanmoran: That's totally fine. Bash will see a path like `/some-org/some-repo//some-dir` and convert that to `/some-org/some-repo/some-dir`.

PR #110, opened 2020-07-23
Updates upstream Mergify config

thitch97:
This adds an upstream configuration file now that we have decided on a preliminary workflow for auto-merging dependabot updates with Mergify.

**Things to note:**
- The Mergify bot will need to be added to the list of users allowed to push to the base branch on each repo. To do this, navigate to the 'Branches' tab in the repo settings and select **"Restrict who can push to matching branches"**. Add `mergify` and the relevant list of maintainers for that repo to the allowed list.

- The `strict rebase` solution we have opted for has potential [drawbacks](https://doc.mergify.io/actions.html#strict-rebase) that may arise during our testing.
- The `bot_account` field specifies the account that will be used by Mergify when rebasing branches as it will need authorization.

### DIFF
--- a/implementation/.mergify/config.yml
+++ b/implementation/.mergify/config.yml
@@ -1,0 +1,12 @@
+pull_request_rules:
+  - name: Auto merge pull requests from dependabot
+    conditions:
+      - author=dependabot[bot]
+      - status-success=Unit Tests
+      - status-success=Integration Tests
+    actions:
+      merge:
+        strict: smart
+        method: rebase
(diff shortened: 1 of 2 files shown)


ryanmoran:
Are we not going to also do this for language-family buildpacks? I don't see why we wouldn't.

thitch97:
> Are we not going to also do this for language-family buildpacks? I don't see why we wouldn't.

I'll update this PR with those changes

PR #122, opened 2020-07-31
scripts: script to validate repo settings

arjun024:
This script can be used to document recommended repo settings
for paketo cnbs.
Script validates settings of the given repo.
Only branch protection rules are validated in this commit.
More rules can be added later.

Signed-off-by: Arjun Sreedharan <asreedharan@vmware.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,6 +7,9 @@ CNBs.
 
 Run [scripts/sanity.sh](scripts/sanity.sh) to see if the changes you made to this repo are valid.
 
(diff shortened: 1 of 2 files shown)


arjun024:
@sophiewigmore Not one that I'm aware of. You need to establish you have the authorization to read the settings of the repo. You can use any token that has admin level access

sophiewigmore:
@arjun024 we are currently in the process of renaming our master branches `main`, and it looks like the script is hardcoded to use `master`

arjun024:
@sophiewigmore Thanks. Switched to main since all paketo cnbs now use main

PR #142, opened 2020-08-17
Adds usage to all scripts

### LINKED ISSUE #79: Add usage flag to package scripts
Could we consider adding in a usage (`-h`) flag to our package scripts so that users can see what options are supported in packaging a buildpack (caching, etc.)?

ryanmoran:
Resolves #79

### DIFF
--- a/scripts/repo_rules.sh
+++ b/scripts/repo_rules.sh
@@ -86,6 +86,7 @@ OPTIONS
   --help  -h         prints the command usage
   --repo <repo>      name of the GitHub repository to check in the form <org>/<name>
   --token <token>    GitHub token used to check the repository
+  --verbose          Print the JSON returned from the API
 USAGE
 }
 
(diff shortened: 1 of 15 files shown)


scripts/repo_rules.sh
```diff
@@ -1,94 +1,260 @@
...
+repo_rules.sh --repo <repo> --token <token> [OPTIONS]
+
+Validates branch protection rules for a GitHub repository.
+
+OPTIONS
+  --branch <branch>  branch to check for protection rules (default: main)
+  --help  -h         prints the command usage
+  --repo <repo>      name of the GitHub repository to check in the form <org>/<name>
+  --token <token>    GitHub token used to check the repository
+USAGE
```
sophiewigmore, line 89: Does --verbose need to be one of the options listed in the --help output also?

ryanmoran: Ah. Yeah. I missed that.

PR #149, opened 2020-08-28
merge push and dispatch workflows

joshzarrabi:
- we only want to send dispatch updates after successfully pushing to
gcr.
- right now, these both run in parallel, and so if we fail to push to
gcr but send the dispatch we break the language family


implementation/.github/workflows/push-buildpackage-and-send-dependency-update.yml
```diff
@@ -1,21 +1,39 @@
-# This workflow notifies language families on release of impl cnbs
-# Requires the following:
-# - secret PAKETO_BOT_GITHUB_TOKEN -> token with permissions to send dispatch to language family repos
-# - file .github/language-family-membership listing target <org>/<repo> per line
```
sophiewigmore, line 4: @joshzarrabi Don't we need to still include these comments so that whoever is creating an implementation buildpack remembers to add the language-family-membership file and the dispatch secret?

joshzarrabi: Yup! Thanks

sophiewigmore (approved):

PR #154, opened 2020-09-11
Make unit/int tests exit non-zero when tests are focused

arjun024:
These tests are used to validate pull requests and it's
very easy to overlook focused tests and assume the change
in PR satisfies the entire test suite.

Signed-off-by: Arjun Sreedharan <asreedharan@vmware.com>

### DIFF
--- a/implementation/scripts/.util/print.sh
+++ b/implementation/scripts/.util/print.sh
@@ -36,7 +36,8 @@ function util::print::success() {
   reset="\033[0;39m"
 
   echo -e "${green}${message}${reset}" >&2
-  exit 0
+  exitcode="${2:-0}"
+  exit "${exitcode}"
 }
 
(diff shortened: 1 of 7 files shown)


sophiewigmore:
@arjun024 Looks good to me! Although, I wonder if the 197 exit code / printing "Detected Focused Test" will be overlooked since the tests still print "succeeded". What do you think?

arjun024:
@sophiewigmore IMO that should be okay. A machine/CI would be able to fail on focused test from the exit code, and at the same time allow casual users to focus tests and see if it succeeds or not. Simulates the behavior of frameworks like ginkgo

sophiewigmore:
That's true -- it lets users test their code. I'm good with that! @arjun024

sophiewigmore (approved):

PR #156, opened 2020-09-21
actions: New generate-builder-toml action

arjun024:
This action generates a builder.toml[1] given an order.toml, stack id and
stack images. All buildpack images, lifecycle-image and build-image are
tagged with their exact latest versions in the generated builder.toml, but
the run-image and run-image-mirrors are tagged with the provided
stack-image-tag
[1] https://buildpacks.io/docs/reference/config/builder-config

The following is an example workflow using this action:

```yaml
- name: Generate builder.toml
  id: generate-builder-toml
  uses: <this-action>
  with:
    stack: "org.cloudfoundry.stacks.cflinuxfs3"
    build-image: "index.docker.io/paketobuildpacks/build"
    run-image: "index.docker.io/paketobuildpacks/run"
    run-image-mirrors: "gcr.io/paketo-buildpacks/run"
    stack-image-tag: "full-cnb-cf"
    order-file: ./order.toml
    registry-server: "gcr.io"

- name: Print
  run: |
    echo '${{ steps.generate-builder-toml.outputs.builder_toml }}' > builder.toml
    cat builder.toml
```

Co-authored-by: Arjun Sreedharan <asreedharan@vmware.com>


arjun024:
For the record: this action implements part of [Builder RFC 0002](https://github.com/paketo-buildpacks/builder/pull/36).

ryanmoran (approved):

PR #162, opened 2020-09-23
Refactors builder.toml generation action

ryanmoran:
- uses more functional patterns to increase testability
- replaces custom image reference parsing with external library


actions/builder/update/entrypoint/'
```diff
@@ -0,0 +1,311 @@
+package main_test
```
sophiewigmore, line 1: I'm a bit confused about the name of this file being `'` is there a reason for this or is it a typo?

actions/builder/update/entrypoint/main.go
```diff
@@ -0,0 +1,66 @@
...
+
+	if options.BuilderFile == "" {
+		fail(errors.New(`missing required input "builder-file"`))
+	}
+
+	if options.RegistryServer == "" {
+		fail(errors.New(`missing required input "registry-server"`))
+	}
+
+	builder, err := internal.ParseBuilderFile(options.BuilderFile)
```
sophiewigmore, line 32: Will this work the first time it runs when there is no builder.toml file yet?

ryanmoran: The assumption is that the `builder.toml` is checked in from the very beginning. The first time, a human will have created a `builder.toml` by hand.

actions/builder/update/entrypoint/'
ryanmoran: Yeah. Looks like this got checked in accidentally.

actions/builder/update/entrypoint/main.go
sophiewigmore: Sounds good, I didn't realize that was the assumption

sophiewigmore (approved):

PR #164, opened 2020-09-29
point to paketo-buildpacks org for ruby

joshzarrabi:


ryanmoran (approved):

.github/data/implementation-cnbs
```diff
@@ -1,3 +1,6 @@
+paketo-buildpacks/bootstrapper
```
sophiewigmore, line 1: @joshzarrabi is bootstrapper supposed to be included in this PR?

joshzarrabi: Yeah we use it keep the template buildpack updated. In any case I didn't add it in this PR I just sorted all of the buildpacks.

sophiewigmore: @joshzarrabi Oh, I meant that it lives in `paketo-community/bootstrapper` instead of `paketo-buildpacks` so it might have been accidentally updated to be in the wrong org.

joshzarrabi: Oh yeah! Good catch!

sophiewigmore (approved):

PR #180, opened 2020-10-26
Calculate time between last PR commit and merge

fg-j:
Rough implementation computing the desired metrics.

Not yet implemented: the ability to include/exclude repos by name


ryanmoran:
Is this ready for review, or should this be a draft?

floragj:
@ryanmoran The current implementation does what is outlined in the acceptance criteria, but what it does could be further refined in a few ways we identified along the way of implementing a first iteration. For instance:
- including/excluding repos by name (so we can remove Java repos if we want). 
- doing something more clever to filter out PRs that sat un-reviewed for a long time

I'm happy to hear feedback from folks about whether we want those and/or other refinements, or if we're good with this initial implementation.

sophiewigmore:
Just realized we should probably add some error statement if the `PAKETO_GITHUB_TOKEN` isn't set so the user knows to set it

floragj:
I changed it up and added concurrency! Also added a check for the github token being set before it's off to the races.

Unfortunately, it's really easy to hit the [github API abuse rate limit](https://developer.github.com/v3/#abuse-rate-limits) making concurrent requests, so the efficiency gains may be purely intellectual/theoretical here.

Not yet added: Reasonable error handling. Still figuring out how to do that well with channels.

sophiewigmore:
Looks pretty good to me, although I think Ryan should probably leave his review as well because I'm not entirely familiar with channels. 
I'm curious if you ran into the API abuse rate limit while testing this out locally?

floragj:
Yeah, I saw it locally. It seems like it's easy to trigger the rate limit if you're making concurrent requests with the same auth token. They expressly indicate that concurrent requests isn't something they want you to do :'(

joshzarrabi:
@fg-j This looks cool! Have you thought about adding tests, we usually have them for our go in this repo.

ryanmoran:
When you run this, are you using your own PAT or the Paketo bot token?

ryanmoran:
I was wondering because I tried with my own token and haven't hit any sort of concurrency limit. I wonder if they are detecting requests from different hosts and you could run into that if you are running the script and an action on a repo is also running.

fg-j:
In the new implementation changes, I've set the default number of concurrent workers to 1 -- @ryanmoran try upping that with your own or the Paketo token and see if you hit the abuse limit. I still saw the same problem when using my PA.

scripts/metrics/internal/init_test.go
```diff
@@ -0,0 +1,15 @@
...
+
+import (
+	"testing"
+
+	"github.com/sclevine/spec"
+	"github.com/sclevine/spec/report"
+)
+
+func TestInternal(t *testing.T) {
+	suite := spec.New("scripts/metrics", spec.Report(report.Terminal{}))
```
ryanmoran, line 11:
```suggestion
	suite := spec.New("scripts/metrics/internal", spec.Report(report.Terminal{}))
```

scripts/metrics/internal/pull_request.go
```diff
@@ -0,0 +1,130 @@
...
+	"math"
+	"net/http"
+	"net/url"
+	"os"
+	"sort"
+	"strings"
+	"time"
+)
+
+const dateLayout string = "2006-01-02T15:04:05Z"
```
ryanmoran, line 16: Is this not just `time.RFC3339`?

scripts/metrics/internal/init_test.go
```diff
@@ -0,0 +1,15 @@
...
+	suite("PullRequest", testPullRequest)
+	// suite("Repository", testRepository)
```
ryanmoran, line 13: Is this a test suite you plan to write?

ryanmoran (approved):

PR #185, opened 2020-11-03
Updates github-config on a schedule

ryanmoran:
Following the implementation of updating dependencies on a schedule that was developed in the builder repos, this changes the updating of the shared github-config to a scheduled workflow.

### DIFF
--- a/implementation/.github/workflows/update-github-config.yml
+++ b/implementation/.github/workflows/update-github-config.yml
@@ -1,8 +1,8 @@
-name: Handle dispatch from github-config
+name: Update shared github-config
 
 on:
-  repository_dispatch:
-    types: working-dir-update
+  schedule:
+  - cron: '*/15 * * * *'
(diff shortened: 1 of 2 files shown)


sophiewigmore:
@ryanmoran I'm not sure that I understand the benefit of having all of our repos check for updates every 15 minutes over just handling dispatches when a change happens. Could you provide some more insight? Also @joshzarrabi thoughts?

joshzarrabi:
@sophiewigmore This is nice because then the github config repo doesn't need to know anything about the downstream repos. So, if we were to add. a new buildpack that needs to get github-cofig updates we won't have to remember to add it to the lists in `.github/data`. So I believe that after this PR is merged we should also get rid of `github-config/.github/workflows/notify-{implementation,language-family}.yml`.

joshzarrabi (approved):

PR #192, opened 2020-11-09
integration.sh: flag to dump test time variables to a file

arjun024:
This may be stored with the buildpack or in the release notes if desired.

See example that uses the optional --dump flag:

```sh
$ ./scripts/integration.sh --dump /tmp/dump.json
$ cat /tmp/dump.json
{
  "builder": "index.docker.io/paketobuildpacks/builder:base@sha256:bae0a6b7ec5d51867eb49ac598775bb5fa8788bba458d2edfda314611be554a8",
  "run": "index.docker.io/paketobuildpacks/run:base-cnb@sha256:155ee4b4389fa88b8b9bd5444601417218b3da44e491b1aad8c8b6c2365e7909",
  "lifecycle": "index.docker.io/buildpacksio/lifecycle:0.9.3@sha256:bc253af2edf1577717618cb3a95f0f16bb18fc9e804efbcc1b85f657d931a757",
  "pack": "0.14.2+git-0fd189d.build-1450"
}
```

### DIFF
--- a/implementation/scripts/.util/tools.sh
+++ b/implementation/scripts/.util/tools.sh
@@ -133,6 +133,16 @@ function util::tools::packager::install () {
     fi
 }
 
+function util::tools::image::reference() {
+  local image digest
+  image="${1}"
+  digest="$(
+    docker inspect --format='{{index .RepoDigests 0}}' "${image}" \
+      | cut -d '@' -f 2
+  )"
+  echo "${image}@${digest}"
+}
+
 function util::tools::tests::checkfocus() {
   testout="${1}"
   if grep -q 'Focused: [1-9]' "${testout}"; then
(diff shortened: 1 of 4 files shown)


sophiewigmore:
I like this idea. If we wanted to attach the output file to our github workflows, is that a change we could add in this PR as well?

fg-j:
This seems generally helpful for reproducibility. I'm having trouble thinking of a specific use-case or pain point this addresses. Is there a recent hard-to-reproduce bug that this is in response to?

sophiewigmore:
@fg-j correct me if I'm wrong @arjun024 but I'm pretty sure we've had a few cases now where our integration tests are failing in GHA,  but we can't easily reproduce the error locally. Having a file that calls out exactly which versions of things are being used would make it simpler for us to repro issues. Also it's nice to have the history in our workflows of what versions of the builder are used

arjun024:
@fg-j This PR doesn't really do anything special. The integration test already prints out these variables but they are scattered throughout. This PR essentially pretty-prints them to a single file which may be stored.
If you checkout [pipenv v0.0.115](https://github.com/paketo-community/pipenv/tree/v0.0.115) and run the integration tests, it fails now but it used to succeed when the release was made. Someone debugging this failure has to find the action log of the integration test that passed, and scrap out the info. If the consumer of the test script had decided to store the file in the release for example, they can access this info easily.

ryanmoran:
Do we have a plan for what we want to do with this data? I'd be much more inclined to merge this if I could understand where and how this data would be accessible to us so that we could use it effectively.

arjun024:
Closing this out for now. Will consider resuming work on this at a later time.

PR #203, opened 2020-12-03
Use pack inspect-builder to get lifecyle and run image details

ryanmoran:
## Summary
The `pack inspect-builder` command will now give us the same details we were using `docker inspect` for. Let's use it as it makes this code a bit simpler.

Please confirm the following:


sophiewigmore (approved):

sophiewigmore:
good idea!

PR #209, opened 2020-12-14
Converts buildpack/update to composite action

ryanmoran:
## Summary
Converts the `buildpack/update` action to a `composite` action type. This allows us to do things like login to a private registry before running the action, and then have that login accessible to the action.

Please confirm the following:


sophiewigmore:
@ryanmoran woah this is awesome. Really clears up the code

actions/buildpack/update/Dockerfile
```diff
@@ -1,20 +0,0 @@
-FROM alpine
-
-RUN apk add \
-    bash \
-    curl \
-  && rm -rf /var/cache/apk/*
-
-RUN mkdir /lib64 && ln -s /lib/libc.musl-x86_64.so.1 /lib64/ld-linux-x86-64.so.2
```
sophiewigmore, line 8: is none of this set up needed now?

sophiewigmore: Update - this isn't needed since the action doesn't run in a docker container now.

sophiewigmore (approved):

PR #214, opened 2021-01-14
Allows automerging PRs in the 'unstable' state

ryanmoran:
## Summary

As outlined in an [issue on octokit](https://github.com/octokit/octokit.net/issues/1763), the `unstable` state is defined as "Failing/pending commit status that is not part of the required status checks. Merging is allowed."

## Use Cases

Running the "Auto Merge" workflow itself induces an `unstable` state as it is a pending check that is not part of the required status checks. This is different than the `blocked` state which is induced by running one of the required status checks. Either way, a merge will only be possible if all of the required checks have passed and so it should be safe to allow the "Auto Merge" workflow to attempt to merge the PR given an `unstable` state.

## Checklist

### DIFF
--- a/implementation/.github/workflows/auto-merge.yml
+++ b/implementation/.github/workflows/auto-merge.yml
@@ -30,7 +30,7 @@ jobs:
         echo "::set-output name=mergeable_state::$(echo "${payload}" | jq -r -c .mergeable_state)"
 
     - name: Merge
-      if: ${{ steps.pull_request.outputs.mergeable_state == 'clean' }}
+      if: ${{ steps.pull_request.outputs.mergeable_state == 'clean' || steps.pull_request.outputs.mergeable_state == 'unstable' }}
       uses: paketo-buildpacks/github-config/actions/pull-request/merge@main
       with:
         user: paketo-bot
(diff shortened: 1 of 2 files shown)


implementation/.github/workflows/auto-merge.yml
```diff
@@ -30,7 +30,7 @@ jobs:
         echo "::set-output name=mergeable_state::$(echo "${payload}" | jq -r -c .mergeable_state)"
 
     - name: Merge
-      if: ${{ steps.pull_request.outputs.mergeable_state == 'clean' }}
+      if: ${{ steps.pull_request.outputs.mergeable_state == 'clean' || steps.pull_request.outputs.mergeable_state == 'unstable' }}
```
sophiewigmore, line 33: are there cases where 'clean' can still happen?

ryanmoran: I think the issue here is that the GitHub API is not consistent in what it returns. I believe there is some sort of race condition wherein sometimes it will return `clean` and other times it returns `unstable`. Both should be acceptable.

sophiewigmore: ahhh ok that's fair then

sophiewigmore (approved):

PR #219, opened 2021-02-02
Check human commits

thitch97:
## Summary

Adds a step to the Test Pull Request workflow to check for human commits on a bot-opened PR.

## Use Cases

This helps to lessen the possibility of bot-opened PRs being commandeered by preventing automatic merging when commits with human authors are present. 

## Checklist


sophiewigmore:
I'm surprised there isn't a way to get all of the commits on a PR without having to curl the repo/PR/use a token. Did you look into something like that?

sophiewigmore (approved):
This looks good to me. I'm waiting to merge until @ryanmoran can take a look too, incase he has any opinions on this

fg-j:
Note: before merge, we want to accept the usernames of bot accounts as an input to the action. (This will make the action more portable.)

fg-j (changes requested):
We're going to make the set of bot accounts to allow an input parameter to the action.

sophiewigmore:
@fg-j want to make this a draft until it's done?

sophiewigmore (approved):
looks good to me. I'd still like @ryanmoran to take a look though as he has a lot of experience with these actions

actions/pull-request/check/entrypoint
```diff
@@ -0,0 +1,81 @@
...
+
+      if [[ ${humanCommit} -eq 0 ]]; then
+        echo "human committer: " "${author}"
+        return 0
+      fi
+  done < <(curl "https://api.github.com/repos/${repo}/pulls/${number}/commits" \
+    --silent \
+    --location \
+    --header \
+    "Authorization: token ${token}" | jq -r '.[] | .commit.author.name, .commit.committer.name')
```
ryanmoran, line 76: Why are we using the `name` fields on the commit rather than the `login` fields under `committer` and `author`?

actions/pull-request/check/entrypoint
```diff
@@ -0,0 +1,81 @@
...
+
+  return 1
+}
```
ryanmoran, line 79: I've tried to simplify the bash used here so that it might be a bit easier to reason about. It ends up being more code, but hopefully uses simpler constructs.

```suggestion
function rules::has_human_commits() {
  local token repo number
  token="${1}"
  repo="${2}"
  number="${3}"

  IFS="," read -r -a bots <<< "${4}"
  IFS=" " read -r -a committers <<< "$(
    curl "https://api.github.com/repos/${repo}/pulls/${number}/commits" \
      --silent \
      --location \
      --header "Authorization: token ${token}" \
      | jq -r '.[] | .author.login, .committer.login' | xargs
  )"
  declare -a unknown

  for committer in "${committers[@]}"; do
    # check to see if this login belongs to a bot
    for bot in "${bots[@]}"; do
      if [[ "${committer}" == "${bot}" ]]; then
        continue 2
      fi
    done

    # check to see if this login is on the list of unknowns
    if [[ "${#unknown[@]}" != 0 ]]; then
      for login in "${unknown[@]}"; do
        if [[ "${committer}" == "${login}" ]]; then
          continue 2;
        fi
      done
    fi

    # add this committer to the list of unknowns
    unknown+=("${committer}")
  done

  # if the length of unknown logins is 0, fail the check
  if [[ "${#unknown[@]}" == 0 ]]; then
    return 1
  fi

  echo "human committers: ${unknown[*]}"
  return 0
}
```

implementation/.github/workflows/test-pull-request.yml
```diff
@@ -42,15 +42,26 @@ jobs:
...
     needs: integration
     steps:
+    - name: Check for Human Commits
+      id: human-commits
+      uses: paketo-buildpacks/github-config/actions/pull-request/check@main
+      with:
+        token: ${{ secrets.PAKETO_BOT_REVIEWER_GITHUB_TOKEN }}
+        repo: ${{ github.repository }}
+        number: ${{ github.event.number }}
+        bots: "dependabot[bot],paketo-bot,GitHub"
```
ryanmoran, line 57: Why are these different than the defaults?

actions/pull-request/check/entrypoint
thitch97: We have discovered that it is possible for the `login` fields under `committer` and `author` to contain `null` in some cases ([example](https://github.com/paketo-buildpacks/dotnet-core-runtime/pull/138/commits/083c0227915c313d15065bbd6209af2aaf79c71a)). We considered it safer and less confusing to use the `name` field rather than perhaps adding `null` as an element in `bots`.

actions/pull-request/check/action.yml
```diff
@@ -0,0 +1,32 @@
...
+  repo:
+    description: 'Github repository'
+    required: true
+  number:
+    description: 'PR number'
+    required: true
+  bots:
+    description: 'A comma-separated list of bot usernames from which to allow commits.'
+    required: true
+    default: 'dependabot[bot],web-flow'
```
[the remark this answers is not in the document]
thitch97: I believe this was an oversight on our part as these values are more consistent with what the `login` field would contain in any case. We can correct this.

implementation/.github/workflows/test-pull-request.yml
thitch97: Related to [this comment](https://github.com/paketo-buildpacks/github-config/pull/219/files#r570508444) I believe.

actions/pull-request/check/entrypoint
ryanmoran: I think this is caused by us using the wrong email when creating commits here: https://github.com/paketo-buildpacks/github-config/blob/509aa0b6e02d46d1a2430e7a2cc92a53335aee7b/actions/sync/entrypoint#L144. I'm going to fix that. I worry that if we use the name of the committer, there isn't anything stopping someone from changing their commit signature name to `paketo-bot` and getting past this check. We'd also want to ensure the commits that come from these bots are signed. That's the only way you can truly verify that the committer is who they say they are.

actions/pull-request/check-commits-unverified/Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+
+ARG gh_version=0.11.1
+RUN mkdir -p "/bin" \
+      && export PATH="${PATH}:/bin" \
+      && curl "https://github.com/cli/cli/releases/download/v${gh_version}/gh_${gh_version}_linux_amd64.tar.gz" \
+        --silent \
+        --location \
+        --output "/tmp/gh.tgz" \
+      && tar xzf "/tmp/gh.tgz" --strip-components 1 -C / \
+      && chmod +x "/bin/gh"
```
ryanmoran, line 20: Doesn't look like the entrypoint uses `gh`, and so this could probably be removed.

actions/pull-request/check-commits-unverified/Dockerfile
```diff
@@ -0,0 +1,23 @@
+FROM alpine
+
+RUN apk add \
+    bash \
+    curl \
+    git \
```
ryanmoran, line 6: You can probably remove `git`. Don't see how it gets used.

actions/pull-request/check/Dockerfile
```diff
@@ -0,0 +1,23 @@
+FROM alpine
+
+RUN apk add \
+    bash \
+    curl \
+    git \
```
ryanmoran, line 6: You can probably remove `git` here. Doesn't look used in the `entrypoint`.

actions/pull-request/check/Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+
+ARG gh_version=0.11.1
+RUN mkdir -p "/bin" \
+      && export PATH="${PATH}:/bin" \
+      && curl "https://github.com/cli/cli/releases/download/v${gh_version}/gh_${gh_version}_linux_amd64.tar.gz" \
+        --silent \
+        --location \
+        --output "/tmp/gh.tgz" \
+      && tar xzf "/tmp/gh.tgz" --strip-components 1 -C / \
+      && chmod +x "/bin/gh"
```
ryanmoran, line 20: Doesn't look like `gh` is used in the `entrypoint`.

fg-j:
Cleaned up the Dockerfiles and fixed a small bug in #224. Merging that PR will update this branch.

ryanmoran:
This looks good on its own, but I have another question. We've introduced this new check to confirm that the commits made by bots are verified. I think that is great as it will truly ensure we are only looking at commits that come from these bots. However, we aren't currently signing any of the commits we produce with our actions using `paketo-bot`. This means that all of the PRs opened by that bot are immediately going to fail this new check.

What is the plan for this? Is there going to be another PR that adds commit signing functionality to allow us to create verified commits using the `paketo-bot`?

fg-j:
@ryanmoran That was my thought, yes. We can spin that out into a separate chore. @thitch97 and I found some examples of folks setting up GPG keys in github actions. (e.g. https://github.com/marketplace/actions/import-gpg). In the interest of "test driving" the bot signing, I was hoping to see this check prevent some auto-merges before we set up the bot signing.

fg-j:
pinging @ryanmoran any further changes needed on this?

ryanmoran (approved):

PR #220, opened 2021-02-03
Use skopeo login instead of docker login when pushing

ryanmoran:
## Summary

Switching to `skopeo login` seems like it may fix these push failures we've seen in our workflows: https://github.com/paketo-buildpacks/dotnet-core-runtime/runs/1817049406?check_suite_focus=true.

## Checklist


sophiewigmore:
Do you have any idea why using `skopeo login` fixes those failures? I'm not against using it, but I'd like to understand the root of the issue

sophiewigmore (approved):

PR #223, opened 2021-02-05
Refactors sync action to only perform rsync

ryanmoran:
## Summary

The sync action can be reduced to simply performing the `rsync` of the shared files between the `github-config` repo and the buildpack repo. We can reuse all of the existing `pull-request/*` actions to replace everything else that this workflow needs.

## Checklist


sophiewigmore (approved):
good idea

PR #225, opened 2021-02-11
use jam update-builder for builder update action

fg-j:
## Summary

Uses the new `jam update-builder` command for the builder/update action.
## Use Cases

This action is used in our builder repos (e.g. [github.com/paketo-buildpacks/base-builder](https://github.com/paketo-buildpacks/base-builder))

## Checklist


fg-j:
Note: Merging this will break builders' current automations because it slightly changes the action's API. (The previous version of the action produces the builder.toml as an output, where this one just modifies the builder.toml in place.) PRs should be opened to update the builders' repos as soon as this merges.

sophiewigmore (approved):

sophiewigmore:
Looks good to me

PR #231, opened 2021-02-19
Delete .github/data directory

ForestEckhardt:
As far as I can tell this data is no longer needed because we do not push updates anymore.

## Checklist


ForestEckhardt:
Is `sanity.sh` something that is still needed?

ryanmoran (approved):

ryanmoran:
It looks like it is still used in the `clone-all-repos.sh` script: https://github.com/paketo-buildpacks/github-config/blob/c1cc64efd54db74f7df413885e61460817858fc2/scripts/clone-all-repos.sh#L46-L57

ForestEckhardt:
Ok. In that case I feel like we should move it out of the workflow at the very least. I will close this PR for now then though.

PR #237, opened 2021-02-23
sign commits in workflows that make paketo bot commits

fg-j:
Signed-off-by: Forest Eckhardt <feckhardt@pivotal.io>

## Summary

Now, bot PRs opened by paketo-bot will be verified. These PRs will pass our check for unverified commits and should automerge (once again).

## Checklist

### DIFF
--- a/builder/.github/workflows/update-builder-toml.yml
+++ b/builder/.github/workflows/update-builder-toml.yml
@@ -27,6 +27,8 @@ jobs:
       with:
         message: "Update builder.toml"
         pathspec: "builder.toml"
+        keyid: ${{ secrets.PAKETO_BOT_GPG_SIGNING_KEY_ID }}
+        key: ${{ secrets.PAKETO_BOT_GPG_SIGNING_KEY }}
 
     - name: Git push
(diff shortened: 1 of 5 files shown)


sophiewigmore:
@fg-j Why does it need both `keyid` and `key`?

fg-j:
When you want to tell `git` about a GPG key, you do it [like this](https://github.com/paketo-buildpacks/github-config/blob/341e54eb1afa3ef2c1942b9ac8532ffc91421220/actions/pull-request/create-commit/entrypoint#L45), which requires the key ID. The key itself needs to be placed on the filesystem and [imported by the gpg utility](https://github.com/paketo-buildpacks/github-config/blob/341e54eb1afa3ef2c1942b9ac8532ffc91421220/actions/pull-request/create-commit/entrypoint#L43) before you can use the key for signing.

See [github documentation](https://docs.github.com/en/github/authenticating-to-github/telling-git-about-your-signing-key) about using GPG keys with git.

ryanmoran (approved):

PR #238, opened 2021-02-23
fixes create-commit action to install tool for signing

fg-j:
also adds 'quiet' config to gpg key import

## Use Cases

Fixes a bug where commits can't be signed because `gpg` isn't installed and incorrect flag used for `base64`. Also configures `gpg` to install the key quietly.

## Checklist


fg-j:
See https://github.com/paketo-buildpacks/tiny-builder/runs/1965204519?check_suite_focus=true for an example of how this is currently failing.

ryanmoran (approved):

PR #249, opened 2021-03-16
Updates gh in actions to 1.7.0

ryanmoran:
- removes extra checks in approve action that are unneeded
- simplifies action APIs for actions using newer gh

## Summary

We have been using a rather old version of the `gh` CLI. This updates the actions that use `gh` to `v1.7.0`. Updating allows us to streamline the API for some of these actions to drop certain fields like `user` because the authentication can be performed with only a token.

Additionally, this includes the removal of a check in the `pull-request/approve` action that asserts bot PRs were opened by either `paketo-bot` or `dependabot[bot]`. This check is unnecessary because that is already checked earlier in the workflows that use this action. It also makes it easier to use this action outside of Paketo buildpacks.

## Checklist

### DIFF
--- a/actions/pull-request/approve/Dockerfile
+++ b/actions/pull-request/approve/Dockerfile
@@ -8,7 +8,7 @@ RUN apk add \
 
 RUN mkdir /lib64 && ln -s /lib/libc.musl-x86_64.so.1 /lib64/ld-linux-x86-64.so.2
 
-ARG gh_version=0.11.1
+ARG gh_version=1.7.0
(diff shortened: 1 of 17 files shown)


joshzarrabi (approved):
Looks good, would there be a way to make sure we don't fall so far behind in the future?

ryanmoran:
I think automating this will be hard and I'm not sure that the consequences of falling behind are that bad. I only updated them because I wanted a new feature, not because anything had broken.

PR #260, opened 2021-04-20
Adds very basic changelog to release notes

### LINKED ISSUE #259: Release Note Automation
Some of our repositories have changelogs in their release notes (check out [packit's release notes](https://github.com/paketo-buildpacks/packit/releases/tag/v0.10.1) for example). This information is helpful for users, but can be time consuming for maintainers. It's also not standardized across our buildpacks and other repos.

We should create some kind of automation that will auto-generate a changelog in the release notes when the `create-draft-release` workflow runs so we can have consistent notes across our community.

ryanmoran:
## Summary

The changelog added here is just a first-pass at something that can involve into a more useful solution. We haven't solidified our use of labels on issues/PRs yet, and so we aren't really in a state to use them to group changelog output yet. Once we do that, it makes sense to start to group these changes. That being said, including something now is better than doing nothing until we come to a consensus.

The output will look like this:
Screen Shot 2021-04-20 at 10 09 17 AM

## Use Cases

Resolves #259

## Checklist

### DIFF
--- a/actions/release/notes/entrypoint
+++ b/actions/release/notes/entrypoint
@@ -1,12 +1,89 @@
 #!/bin/bash
 
+set -e
+set -u
+set -o pipefail
+
 function main() {
+  local token repo
+  while [ "${#}" != 0 ]; do
+    case "${1}" in
+      --token)
+        token="${2}"
+        shift 2
+        ;;
+
+      --repo)
+        repo="${2}"
+        shift 2
+        ;;
+
+      "")
+        shift
+        ;;
+
+      *)
+        echo "unknown argument \"${1}\""
+        exit 1
+    esac
+  done
+
+  if [[ -z "${token}" ]]; then
+    echo "--token is a required flag"
+    exit 1
+  fi
+
+  if [[ -z "${repo}" ]]; then
+    echo "--repo is a required flag"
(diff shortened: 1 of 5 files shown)


actions/release/notes/entrypoint
```diff
@@ -1,13 +1,90 @@
...
+  )"
+
+  if [[ -n "${changes}" ]]; then
+    content="$(
+      echo "${content}"
+      echo
+      echo "### Changes"
+      echo "${changes}"
+    )"
+  fi
```
sophiewigmore, line 86: Would it be possible for us to have some kind of check here to exclude any github-config bump commits?

This is fabulous, but the only part I am concerned about is the `Update github-config @paketo-bot` commits that get added. 

The packit bumps are useful, but to users looking at the release notes, the fact that the github-config got updated doesn't help much with anything. Also, an outside user might not know what that is, and it can make the changelog more difficult to understand.

sophiewigmore (changes requested):

actions/release/notes/entrypoint
ryanmoran: Done.

sophiewigmore (approved):

PR #261, opened 2021-04-21
Uploads workflow event payload after tests for PRs

ryanmoran:
## Summary

Recently, we have been seeing PRs that are opened by Dependabot failing due to a token permissions issue when the Test PR workflows attempt to auto-approve the PR. Our understanding is that this behavior by GitHub is correct, but it breaks our current workflow for these types of PRs. We would like to move to a workflow that will work in all cases.

This PR is the start of an attempt to create a pair of workflows that work in conjunction to give us this outcome. First, the existing portions of the Test PR workflow will execute, then a subsequent workflow that is separate, and with elevated permissions will run to perform the step to auto-approve the bot PR. In order to pass details about the Test PR to the second workflow, we will need to upload that information to the workflow's artifact store. This PR does that for each of the Test PR workflows.

The next steps from this will be a follow-on PR that then creates a second workflow that can download this artifact and act on it to auto-approve the bot PR.

## Checklist

### DIFF
--- a/builder/.github/workflows/test-pull-request.yml
+++ b/builder/.github/workflows/test-pull-request.yml
@@ -22,6 +22,16 @@ jobs:
     - name: Run Smoke Tests
       run: ./scripts/smoke.sh
 
+  upload:
+    name: Upload Workflow Event Payload
+    runs-on: ubuntu-latest
+    steps:
+    - name: Upload Artifact
+      uses: actions/upload-artifact@v2
+      with:
+        name: event-payload
+        path: ${{ github.event_path }}
+
   approve:
     name: Approve Bot PRs
     if: ${{ github.event.pull_request.user.login == 'paketo-bot' || github.event.pull_request.user.login == 'dependabot[bot]' }}
--- a/implementation/.github/workflows/test-pull-request.yml
+++ b/implementation/.github/workflows/test-pull-request.yml
@@ -41,6 +41,16 @@ jobs:
       env:
         GIT_TOKEN: ${{ github.token }}
 
+  upload:
+    name: Upload Workflow Event Payload
+    runs-on: ubuntu-latest
+    steps:
+    - name: Upload Artifact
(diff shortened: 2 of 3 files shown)


sophiewigmore:
@ryanmoran what bits about the workflow are included in the "workflow artifact" that gets uploaded? It's hard for me to imagine what the next workflow will look like without knowing that.

Also, I see that the existent Auto-Approve part is still there right now. Will this be removed in the next PR in favour of that 2nd workflow? Also could you briefly describe how a second workflow could have the "right" permissions that we couldn't set in this workflow?

ryanmoran:
> what bits about the workflow are included in the "workflow artifact" that gets uploaded?

We will just be uploading the event payload. This is the JSON metadata that includes information like which PR triggered the workflow, etc.

> Also, I see that the existent Auto-Approve part is still there right now. Will this be removed in the next PR in favour of that 2nd workflow?

The `approve` job will be moved into its own workflow and setup to execute when the "Test PR" workflow completes successfully. In that step we will download the event payload artifact and then perform the steps to audit the PR to determine if it can be automatically approved. 

> Also could you briefly describe how a second workflow could have the "right" permissions that we couldn't set in this workflow?

The new workflow that will be added as described above will have `write` permissions (which grants access to secrets) and so will be able to approve PRs. If you read through [these docs](https://docs.github.com/en/actions/reference/authentication-in-a-workflow#permissions-for-the-github_token), you can see a more detailed explanation of the permissions granted for different types of workflows.

In short, we are going to be moving to the prescribed workflow structure outlined in this blog post: https://securitylab.github.com/research/github-actions-preventing-pwn-requests/. They have an example of the "second" workflow that leverages some gnarly JS code. I'd like to see if we can encapsulate that same type of logic in our own action.

sophiewigmore:
Thank you for the explanation!

sophiewigmore (approved):

PR #269, opened 2021-05-03
docs.sh for language family repos that have docs Hugo modules

fg-j:
## Summary

The script `scripts/docs.sh` installs Hugo and uses it to serve a local version of the Paketo website with the local copy of the language family docs replacing the appropriate Hugo module.

## Use Cases

This allows contributors and code reviewers to quickly preview changes to the language family docs without having to manually clone and interact with the paketo-website repo. **This script will fail to run in repos that don't yet have a `docs/` directory + Hugo module.** Currently, it is only useful for the paketo-website/dotnet-core repo, but will become applicable to others as we expand the use of Hugo modules.

## Checklist


sophiewigmore (changes requested):
@fg-j This looks great! One thing: https://github.com/paketo-buildpacks/github-config/pull/268/ went in right before this PR and it adds a workflow and action to keep our tool versions up to date. Do you think you could pull in those changes here and just add `hugo` to those files so it will also stay up to date?

fg-j:
@sophiewigmore I chose to clone a clean copy into the repo to make minimal assumptions about the user's directory setup. This way, the script just works even if a) you haven't already cloned a copy of paketo-website or b) you have cloned a copy and you have local changes that you want to keep or c) you have already cloned a copy but it's not in a predictable location relative to the buildpack directory.

Also, it means that the script doesn't dump files into some unexpected other place on the filesystem 🙂  . I followed the model of installing binaries into `.bin` rather than plopping them in `/tmp/` or so forth.

fg-j:
@sophiewigmore just want to note that I noticed a small bug in the existing update tools workflow: the implementation and language family buildpacks' tools.json were missing an entry for jam. I added that in.

sophiewigmore (approved):

sophiewigmore:
@fg-j good catch, thank you!

PR #278, opened 2021-05-11
Swaps .packit file references for .libbuildpack

ryanmoran:
## Summary

Now that the majority of the buildpacks are using `packit` as their library, we can remove a bunch of extraneous files by swapping the conditional in the packaging script that determines which tool to use to package the buildpack. Once this is merged and propagates to all of the implementation repos, we can remove all of the `.packit` files from those repos. We can also remove the `.packit` file from the `bootstrapper`.

## Checklist

### DIFF
--- a/implementation/scripts/integration.sh
+++ b/implementation/scripts/integration.sh
@@ -67,7 +67,7 @@ function tools::install() {
   util::tools::jam::install \
     --directory "${BUILDPACKDIR}/.bin"
 
-  if [[ ! -f "${BUILDPACKDIR}/.packit" ]]; then
+  if [[ -f "${BUILDPACKDIR}/.libbuildpack" ]]; then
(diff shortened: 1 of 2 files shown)


sophiewigmore:
What if someone wants to package their buildpack with the `packager` but their buildpack doesn't use `packit` or `libuildpack`? Does this shoehorn them into having a `.libbuildpack` file?

ryanmoran:
That seems like a pretty small, maybe non-existent, group of buildpacks. Are we really trying to focus on providing such a generic mechanism for that? Arguably the same thing could have been said for the existing functionality where a user wants to use `jam pack` but isn't using `packit`. They'd would have needed to provide a `.packit` file.

sophiewigmore (approved):

PR #307, opened 2021-06-21
Adds concurrency group to release workflows

ryanmoran:
## Summary

This should prevent these workflows from running in parallel and ensure we don't accicentally create duplicate releases.

## Checklist


fg-j:
@ryanmoran I noticed that `concurrency` is currently in Beta according to the [GHA docs](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#concurrency). I'm not sure how long it's been in that state. Have we run into problems with duplicate releases so far? Does this cause weird release drafting behaviour? Wondering if it's causing us enough pain to use this Beta feature

ryanmoran:
This is the most recent instance of a duplicate draft release being created: https://paketobuildpacks.slack.com/archives/C01MM2E1CMT/p1624288314001300. I've seen it several times before, but until the `concurrency` feature was added, there wasn't a very easy way to prevent it from recurring.

What is the hesitation with using the feature?

fg-j (approved):
I was hesitant to add a "subject to change" GHA feature  to guard against a hypothetical problem. However, given we've got examples of running into this double-release issue, I'm good with adding the fix.

PR #332, opened 2021-08-18
Bumps Go to 1.17 across all workflows

ryanmoran:
## Summary

All workflows should use Go v1.17 or higher.

## Checklist

### DIFF
--- a/.github/workflows/test-pull-request.yml
+++ b/.github/workflows/test-pull-request.yml
@@ -14,7 +14,7 @@ jobs:
     - name: Setup Go
       uses: actions/setup-go@v1
       with:
-        go-version: 1.16.x
+        go-version: 1.17.x
     - name: Checkout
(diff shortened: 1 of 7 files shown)


fg-j:
I noticed that many of our go-based actions (e.g. dispatch, release/reset-draft, release/create) are using outdated versions of go (as old as 1.14) in their `go.mod` files.  We should probably upgrade those to 1.17 as well.

ryanmoran:
The value recorded in the `go.mod` file isn't the version of Go to use. It defines the minimum compatible version. If you compile with 1.17, then it uses 1.17, even if the `go.mod` file specifies 1.14. You can check out the docs defining this [here](https://golang.org/doc/modules/gomod-ref#go). Its maybe worth updating at some point, but we shouldn't feel compelled to do it immediately.

ForestEckhardt (approved):

PR #359, opened 2021-10-25
Calculate next release version based on PR labels

fg-j:
## Summary

An action that uses similar logic as the  [release-notes action](https://github.com/paketo-buildpacks/github-config/blob/00ebdc9b7db5ec72c447d7b70c1f3d9f37610eae/actions/release/notes/entrypoint) to identify commits/PRs since the last release on a repo and uses the labels on those commits to determine the correct semver for the next release. This action would take the place of [increment tag](https://github.com/paketo-buildpacks/github-config/blob/00ebdc9b7db5ec72c447d7b70c1f3d9f37610eae/actions/tag/increment-tag/action.yml) in workflows like [create-draft-release](ihttps://github.com/paketo-buildpacks/github-config/blob/00ebdc9b7db5ec72c447d7b70c1f3d9f37610eae/implementation/.github/workflows/create-draft-release.yml#L62).

## Use Cases

This action, combined with some PR-labeling automation, will remove the need for maintainers to manually adjust versions of releases before cutting them.

## Checklist


sophiewigmore:
One last question - if we determine that the automated semver version bump from the automation isn't right for some reason or another, is there a way for us to override the version?  You know that weird state where you want to bump the minor or major version in a GH draft release yourself, and then you have to create the tag in the draft and trigger the `Create Release` job (so the assets get compiled with the right name and the release notes). What happens if we want to manually override that version, but the semver determining logic always overrwrites it?

While this automation is a huge time saver, I think it's important to be able to have a mechanism to override it if needed. From what I can tell that's not possible right now. What do you think?

fg-j:
Really really good point. Let me see what I can do on that.

sophiewigmore:
Happy to discuss some options for this as well

fg-j:
@sophiewigmore I think the manual override should be handled at the workflow level. Have a look at what I've done in #370 to address this.

sophiewigmore (approved):

PR #400, opened 2022-02-09
Autolabel dependabot PRs as patch changes

fg-j:
## Summary

When dependabot PRs go out to our repos, they require a lot of maintainer intervention since they aren't auto-labeled. 

The simplest mitigation I can think of is to automatically label dependabot PRs as patch changes. My reasoning is that dependency bumps that *shouldn't* be labeled as patch changes to the buildpack will cause test failures. The main dependency bump that I slightly worry about autolabeling as patch is packit. But:
- we usually mark unused structs, fields, functions as deprecated, which will cause linter failures
- otherwise, we add typically features that buildpack authors need to take action to use; simply bumping packit doesn't generally add buildpack features.

So I might be convinced that it's fine to label all dependency bumps as patches. Let's hash this question out here.

## Checklist

### DIFF
--- a/actions/pull-request/auto-semver-label/action.yml
+++ b/actions/pull-request/auto-semver-label/action.yml
@@ -24,13 +24,14 @@ runs:
         exit 0
       fi
         echo "::set-output name=status::true"
-  - name: Check Files Changed
+  - name: Check For Auto-labelable Changes
     id: changes
     shell: bash
     run: |
       if [[ "${{ steps.auth.outputs.status }}" == "true" ]]; then
         ${{ github.action_path }}/check-files-changed.sh --repo "${{ github.repository }}" \
         --number "${{ github.event.number }}" \
+        --author "${{ github.event.pull_request.user.login }}" \
         --patchfiles "${GITHUB_WORKSPACE}/.github/.patch_files"
       fi
   - name: Clear Semver Label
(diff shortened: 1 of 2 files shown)


robdimsdale:
The idea of labeling dependency bumps as `semver:patch` makes sense to me. It seems low risk and fairly high value (as the majority of PRs that I manually add labels to are dependency bumps). I am struggling to articulate a plausible scenario that bumping a dependency both introduces a minor/major change (rather than a patch change) and yet also doesn't impact our automated tests.

The best example (and it's pretty weak) that I can come up with is something like a change to the TOML/YAML specification: if the dependencies we use to parse buildpack config files supported this new language specification then we could make the case that it increases the surface area supported by the buildpacks and hence should be a minor change.

sophiewigmore:
+1 Rob. 
The only thing that *might* challenge this use case is that I feel like buildpacks maintainers may not always be aware of new features in packit/other tooling that may streamline the buildpack implementation. Sometimes new features might not necessarily break the linter or tests, but would be ideal for usage. An example would be the [addition of a serve matcher in occam](https://github.com/paketo-buildpacks/occam/pull/75). Being aware of bumps to our tooling lessens the possibility  of new versions/features coming out without maintainers knowing.

fg-j:
@paketo-buildpacks/tooling-maintainers seems like we have 👍  from a good few folks. Any outstanding concerns?

ryanmoran (approved):

PR #413, opened 2022-02-16
Adds a ref-name parameter to the calculate semver action

### LINKED ISSUE #412: Incorrect SemVer Calculation
When cutting a release of `packit` we noticed the the SemVer calculation was incorrect based on the labels given to PRs in that release, which can be seen [here](https://github.com/paketo-buildpacks/packit/runs/5223732512?check_suite_focus=true). After looking through the SemVer calculation action I think that I have found the issue. 
https://github.com/paketo-buildpacks/github-config/blob/820df1fc6631ca994a2126e7e9818cd227e2dff7/actions/tag/calculate-semver/entrypoint/main.go#L145

In this api call we are hard coded to compare against `main` however `packit` no longer has main as its default branch instead having `v2` be the default. We should find a way to either allow this to be changed or determine it in logic.

ForestEckhardt:
Resolves #412 

## Checklist

### DIFF
--- a/actions/tag/calculate-semver/action.yml
+++ b/actions/tag/calculate-semver/action.yml
@@ -21,3 +21,5 @@ runs:
   - ${{ inputs.repo }}
   - "--token"
   - ${{ inputs.token }}
+  - "--ref-name"
+  - ${{ inputs.ref-name }}
--- a/actions/tag/calculate-semver/entrypoint/main.go
+++ b/actions/tag/calculate-semver/entrypoint/main.go
@@ -23,6 +23,7 @@ type Config struct {
 	Endpoint string
 	Repo     string
 	Token    string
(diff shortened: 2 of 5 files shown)


sophiewigmore:
Hi! Can you point me to the docs where you found `github.ref.name` I can only see `github.ref_name` (per https://docs.github.com/en/actions/learn-github-actions/contexts) and want to make sure we have the right thing

ForestEckhardt:
@sophiewigmore Good catch. I got the name wrong give me one moment to change it

> Alternatively it might be nice if the ref name is optional, and defaults to main if not passed in.
I don't think that is necessary seeing as the inputs are tightly controlled within a workflow and I am using a default github context

sophiewigmore (approved):

PR #437, opened 2022-03-15
Put repo cron jobs in concurrency groups

### LINKED ISSUE #436: Can we use the concurrency setting in GitHub workflows?
We recently experienced a large backup in the number of queued workflow runs across our GitHub org. This resulted in a massive slowdown in our ability to get PRs merged, releases cut, dependencies updated, etc.

Can we use the [concurrency](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#concurrency) setting on a workflow to reduce some of the queue pressure for our org?

fg-j:
## Summary

Resolves #436 

## Use Cases

Since I think the main cause of the problem is long backed up queues of recurring cron jobs, I've only added concurrency groups to crons. If actions that run on PRs are also playing a significant role, happy to add more concurrency groups that minimize the number of concurrent tasks running on a given PR.

## Checklist


ForestEckhardt:
Do we want to add `cancel-in-progress: true` as well or do we feel just canceling all pending jobs is sufficient.

ryanmoran:
The `cancel-in-progress` options sounds undesirable. Imagine we have a job that queues every 15 minutes, but when it actually needs to do work it takes 20 minutes to complete. If you `cancel-in-progress`, this job will never actually finish what it needs to get done because it will always kill that in-progress job before it can complete.

ForestEckhardt:
That is sound logic to me. I am all good with this then!

ForestEckhardt (approved):

PR #488, opened 2022-06-24
Retry stack USN polling on 504 error

sophiewigmore:
## Summary

Since setting up release failure notification issues on the new Jammy Tiny stack, we have already seen many instances of failed builds that occur because of intermittent 504 errors when polling for USNs. ([example](https://github.com/paketo-buildpacks/jammy-tiny-stack/issues/20))
 All of these notifications are red herrings, because the issue usually disappears when the job runs again. It's creating a lot of notifications.

This PR just allows for a retry in instances where polling fails do to a 504 error, so we can hopefully get less failures during this step.

## Checklist

### DIFF
--- a/actions/stack/get-usns/entrypoint/go.mod
+++ b/actions/stack/get-usns/entrypoint/go.mod
@@ -7,6 +7,7 @@ require github.com/mmcdole/gofeed v1.1.3
 require (
 	github.com/PuerkitoBio/goquery v1.5.1 // indirect
 	github.com/andybalholm/cascadia v1.1.0 // indirect
+	github.com/cenkalti/backoff/v4 v4.1.3 // indirect
 	github.com/json-iterator/go v1.1.10 // indirect
 	github.com/mmcdole/goxpp v0.0.0-20181012175147-0068e33feabf // indirect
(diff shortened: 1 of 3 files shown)


fg-j:
I’m curious if you considered using an [exponential backoff](https://en.m.wikipedia.org/wiki/Exponential_backoff) rather than a fixed retry interval (e.g. this [golang implementation](https://github.com/cenkalti/backoff)). Might protect us from being rate limited on our retries.

sophiewigmore:
@fg-j I hadn't tried this, interesting idea!  My first thought is that doing an exponential backoff in this scenario feels overkill on the level of the action itself running, because all I want to do is try one more time if a query fails the first time--which seemed to have fully correctional results when used on my local workstation. If anything, it might make more sense if we used that algorithm to figure out how often to run the `create-release` workflow itself, which feels like a bigger change

sophiewigmore:
@fg-j check out the newest commit -is this what you had in mind? I also expanded it to do retries on non-504 errors as well.

fg-j (approved):
I just played around with this and managed to get some 504s to test out the backoff. It seems effective! LGTM

PR #493, opened 2022-07-05
actions: fix apk add

arjun024:
Previous code errors with:
`ERROR: gpg-2.2.35-r4: package mentioned in index not found (try 'apk update')`


fg-j:
I'm not sure if I understand how this change resolves the stated error. I see that error intermittently. 
For instance, on the nginx buildpack, here's it [failing](https://github.com/paketo-buildpacks/nginx/runs/7199489829?check_suite_focus=true) and here it is [passing](https://github.com/paketo-buildpacks/nginx/runs/7200685822?check_suite_focus=true)

arjun024:
@fg-j From what I understand, `--no-cache` option is equivalent to wrapping `apk add` with `apk update` in the beginning and` rm -rf /var/cache/apk/*` in the end. The `update` part makes sure that the index of available packages is always updated before the `add` and thus more robust. Sorry I should have added this in the original PR but I just figured it out :)

fg-j:
Nice. Makes sense to me. Is [this](https://stackoverflow.com/questions/49118579/alpine-dockerfile-advantages-of-no-cache-vs-rm-var-cache-apk) your source? (for posterity)

arjun024:
yes & https://github.com/alpinelinux/apk-tools/commit/c43bfed8deaa0dab47c54db9b8f374853d345a6b

fg-j (approved):

PR #495, opened 2022-07-06
Add retry around querying the stack USN RSS feed

sophiewigmore:
## Summary

In #488 we added a retry around querying Stack USNs on 504 failures. We are seeing an occasional failure still with a related step in the same workflow - querying the RSS feed. I have added similary retry logic when querying the RSS  feed to hopefully cut down on the number of failures we see around this.

Example failure: https://github.com/paketo-buildpacks/jammy-tiny-stack/issues/27

## Checklist

* [X] I have viewed, signed, and submitted the Contributor License Agreement.
* [X] I have linked issue(s) that this PR should close using keywords or the Github UI (See [docs](https://docs.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue))
* [ ] I have added an integration test, if necessary.
* [X] I have reviewed the [styleguide](https://github.com/paketo-buildpacks/community/blob/main/STYLEGUIDE.md) for guidance on my code quality.
* [X] I'm happy with the commit history on this PR (I have rebased/squashed as needed).

PR #496, opened 2022-07-07
[Patch] Add retry around querying the stack USN RSS feed

fg-j:
Adds to #495 
Checks whether the error that comes from feed parsing is a retryable HTTP error or a permanent error. Also, checks the error returned by the retry so execution halts if the poller fails to get the feed.


sophiewigmore:
looks good to me!

PR #500, opened 2022-07-13
Fix unbound builders variable

joshuatcasey:
Language family `integration.sh` could have an unbound variable `builders` when `integration.json` is not present.

Example: https://github.com/paketo-buildpacks/python/runs/7314119664


joshuatcasey:
Closing in favor of #499

Issue #467, opened 2022-05-10
Always update version of pack when a new version is available


robdimsdale:
When there is a new version of `pack` available but I already have a version installed under `./.bin/pack` in an implementation buildpack, these scripts do not update `pack` to the newer version. I think they should.

I think there are valid arguments for not updating it, but I think it is simpler to know that we are always running against the latest version when we run integration tests

I think we would expand the logic in [`.util/tools.sh`](https://github.com/paketo-buildpacks/github-config/blob/81f770ef0fc91714a067712965b9f7dbb1530f9f/language-family/scripts/.util/tools.sh#L96) to run `./.bin/pack version` and compare the versions.

I'm happy to submit a PR if folks agree this is worth doing.

ForestEckhardt:
This idea sounds great to me! I am in favor of the idea.

ryanmoran:
The only thing I would caution is that there have been instances in the past where the `pack` version contained a show-stopping bug that required us to roll back to a stable version. Would we want a flag to disable this auto-update feature?

ForestEckhardt:
@ryanmoran I assumed that we would be just checking against `tools.json` to see that we have the version specified in that file. We could gate the releases that way. But if that is not the intention then I agree with your concern.

robdimsdale:
@ryanmoran i didn't realize `pack` had released bugs/breaking changes of such severity in the past.

I think I would support @ForestEckhardt 's suggestion - always download the version in `tools.json`. That way if there is a bug it is easy to revert that file - both locally and by PRing to this repository.

While I haven't been burned by moving forward too quickly on `pack`, I have encountered issues where I forgot to upgrade my local copy of `pack` which caused issues.

ryanmoran:
That seems fine to me. It wasn't clear that's what was suggested.

PR #501, opened 2022-07-18
Print jam and pack versions

joshuatcasey:
.... So that we have a little extra visibility and reminder if the versions are drifting.


joshuatcasey:
Thanks @robdimsdale I should have checked open issues and seen #467 before creating this.

Part of what I was going for is just visibility that the integration tests are using the local copies of the CLIs instead of the system-install versions. I tend to forget this and get frustrated when an integration test build fails and yet  `$ pack build <>` succeeds.

PR #503, opened 2022-07-20
Test buildpacks against multiple builders

fg-j:
## Summary

This PR updates automation that will allow buildpack authors to test against multiple builders without making any changes to existing buildpack Golang tests. 

`./integration.json` is the source of truth for the default set of builders to test against. If no `builder` or `builders` are provided there, the automation falls back to the existing default, the bionic buildpackless builder.

When running tests locally, `scripts/integration.sh` will sequentially set each builder as the default builder, then run the full battery of integration tests.

It's possible to run tests against a specific subset of builders. Use `scripts/integration.sh --builder uri-of-builder --builder other-uri` to override whatever is provided in `integration.json`.

In CI, Github Actions use the `--builder` flag to run tests against each builder _in separate jobs_. The logic that determines which builders to test against is shared between local and GHA runs. It lives in `scripts/.util/builders.sh`. This will make it easy to change defaults down the road.

It'd be helpful if @paketo-buildpacks/tooling-maintainers and others can play around with this on other buildpack repos to see how it feels. Leave feedback here about bugs, pain points, etc. You can test these out on your own repo by copying the `language-family` or `implementation` files into a repo of the appropriate type.

Here's one example using the Go language family: https://github.com/paketo-buildpacks/go/pull/632
And here's an example with the go-dist implementation buildpack: https://github.com/paketo-buildpacks/go-dist/pull/420

## Checklist


sophiewigmore:
@fg-j I just tested this out with https://github.com/paketo-buildpacks/node-module-bom/pull/87, which has two builders in the `integration.json`. When I ran just `integration.sh`, it looks like the tests only ran with the first builder. I would have expected it to run the test suite twice - once for each builder. Is that assumption correct?

sophiewigmore:
@fg-j this works perfectly now! I was able to test out all of the local testing scenarios:
- Saw tests run against all the builders in the `integration.json`
- Saw tests run against all builders passed in as arguments to the script
- Saw tests run against only the one builder I specified when passed as an argument 

I'm happy with this and would be comfortable using this as our testing strategy in the buildpacks

sophiewigmore:
@fg-j is anything else needed to get this undrafted?

sophiewigmore:
I am good with this! Other @paketo-buildpacks/tooling-maintainers, any thoughts before we merge it in?

sophiewigmore (approved):

ryanmoran (approved):

PR #508, opened 2022-07-22
Add workflow to run 'go get -u' and create a PR.

joshuatcasey:
I've been seeing some `Dependabot` alerts about transitive dependencies that usually seem fixed by running `go get -u`. Let's just keep those up-to-date automatically.

Example `Dependabot` alerts: https://github.com/paketo-buildpacks/cpython/security/dependabot?q=is%3Aclosed

Assumes that implementation and language family repos have a go.mod at the root

### DIFF
--- a/implementation/.github/workflows/go-get-update.yml
+++ b/implementation/.github/workflows/go-get-update.yml
@@ -20,14 +20,14 @@ jobs:
 
       - shell: bash
         run: |
-          go get -u
+          go get -u all
           go mod tidy
 
       - name: Commit
         id: commit
         uses: paketo-buildpacks/github-config/actions/pull-request/create-commit@main
         with:
-          message: "Running 'go get -u'"
+          message: "Running 'go get -u all'"
           pathspec: "."
           keyid: ${{ secrets.PAKETO_BOT_GPG_SIGNING_KEY_ID }}
           key: ${{ secrets.PAKETO_BOT_GPG_SIGNING_KEY }}
@@ -43,5 +43,5 @@ jobs:
         uses: paketo-buildpacks/github-config/actions/pull-request/open@main
         with:
           token: ${{ secrets.PAKETO_BOT_GITHUB_TOKEN }}
-          title: "Running 'go get -u'"
+          title: "Running 'go get -u all'"
           branch: automation/tools/go-get-update
--- a/implementation/.github/workflows/go-get-update.yml
+++ b/implementation/.github/workflows/go-get-update.yml
@@ -1,0 +1,47 @@
+name: Go Get Update
+
+on:
+  schedule:
+    - cron: '0 0 * * *'  # Once a day at midnight
+  workflow_dispatch: {}
+
+jobs:
+  update:
+    name: Go Get Update
+    runs-on: ubuntu-latest
+    steps:
+      - name: Setup Go
+        uses: actions/setup-go@v3
+        with:
+          go-version: 1.18.x
+
+      - name: Checkout
+        uses: actions/checkout@v3
+
+      - shell: bash
+        run: |
+          go get -u
+          go mod tidy
+
+      - name: Commit
+        id: commit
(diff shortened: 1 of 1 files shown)


fg-j:
@joshuatcasey It surprises me that dependabot doesn't handle this already. I dug around a bit in the dependabot source code and found [this code](https://github.com/dependabot/dependabot-core/blob/5b347f6f9ec030d94edf2563df628d58a82ca547/go_modules/lib/dependabot/go_modules/update_checker.rb#L16) which to me is a bit unclear about whether it _can_ or _cannot_ update indirect (aka transitive) dependencies.

joshuatcasey:
> @joshuatcasey It surprises me that dependabot doesn't handle this already. I dug around a bit in the dependabot source code and found [this code](https://github.com/dependabot/dependabot-core/blob/5b347f6f9ec030d94edf2563df628d58a82ca547/go_modules/lib/dependabot/go_modules/update_checker.rb#L16) which to me is a bit unclear about whether it _can_ or _cannot_ update indirect (aka transitive) dependencies.

I'm also surprised at the behavior. The comment you found seems to indicate that `dependabot` *will* update those indirect dependencies that are found in `go.mod`, although I'm not sure I can actually find an example where that's happened. We could probably start including indirect dependencies in `go.mod` as a common practice on our team whether we merge this PR or let Dependabot handle it.

fg-j:
> We could probably start including indirect dependencies in go.mod as a common practice on our team

All the `go.mod` files I have seen in the project include indirect dependencies in a separate `require` block (e.g. [this one](https://github.com/paketo-buildpacks/go-dist/blob/main/go.mod)). What further change are you proposing?

One thing to point out is that PRs created by this automation will _not_ be automatically semver labeled, [since they aren't created by `dependabot`](https://github.com/paketo-buildpacks/github-config/blob/a3bbc96f7d27b74c5c288de8e0f95f31d4e062d7/actions/pull-request/auto-semver-label/check-files-changed.sh) and they will change files that aren't in the [patch allowlist](https://github.com/paketo-buildpacks/github-config/blob/a3bbc96f7d27b74c5c288de8e0f95f31d4e062d7/implementation/.github/.patch_files). I think that's fine for now. If this creates a bunch of excess maintainer toil, we can decrease the update frequency.

As far as next steps, 
- please fix up whatever is failing the YAML linter
- and file an issue against dependabot-core so we can retire this automation at some point in the future

language-family/.github/workflows/go-get-update.yml
```diff
@@ -0,0 +1,47 @@
...
+      uses: actions/setup-go@v3
+      with:
+        go-version: 1.18.x
+
+    - name: Checkout
+      uses: actions/checkout@v3
+
+    - shell: bash
+      run: |
+        go get -u
```
fg-j, line 23: I just tried running this in the Go language family buildpack repo. The result was:
```
go: no package to get in current directory
```

This is because `go get` [fails for packages that don't have a `main()`](https://github.com/golang/go/issues/52179). 

So, I don't think we can/should add this automation for language family buildpacks, since they don't have `main()` functions.

fg-j (changes requested):

language-family/.github/workflows/go-get-update.yml
joshuatcasey: I've removed the language family workflow.

I have a vague memory of running `go get -u` on the `cpython` language family buildpack. Weird. https://github.com/paketo-buildpacks/python/pull/488

ForestEckhardt: I think you may want `go get -u all` instead. Running that on the Go language family buildpack yields results.

joshuatcasey: Nice find @ForestEckhardt, I've updated the PR

ForestEckhardt:
@fg-j Could you take another look at this?

fg-j (approved):

PR #519, opened 2022-08-01
Fix issue-opening action

sophiewigmore:
## Summary

In https://github.com/paketo-buildpacks/dep-server/issues/196, we saw an example of a workflow failure issue being reopened by GHA even though the dependencies of interest were different. This change makes it so that if an issue is reopened and/or commented on, the issue title must be what we expect it to be as well.

## Checklist

* [X] I have linked issue(s) that this PR should close using keywords or the Github UI (See [docs](https://docs.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue))

### DIFF
--- a/actions/issue/file/action.yml
+++ b/actions/issue/file/action.yml
@@ -45,7 +45,6 @@ runs:
 
       gh auth status
 
-      release_issue=""
       if [[ "${{ inputs.comment_if_exists }}" == "true" ]]; then
         if [ -z "${{ inputs.label }}" ]; then
           echo "must provide 'label' as input when 'comment_if_exists' is true"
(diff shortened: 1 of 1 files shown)


fg-j:
Maybe it makes more sense to turn off `comment_if_exists` for the dep_server rep since the failures tend to be more heterogeneous? I think I'd find it confusing if the title of the issue kept changing  - then, previous failures on the issue wouldn't correspond to the title anymore.

sophiewigmore:
@fg-j the implementation here is logically the same thing as adding a label, because the title has the dependency name in it. I like that failures in the same dependency will result in comments on the same issue. 

I would like to change the workflow to stop reopening issue which is also what happened https://github.com/paketo-buildpacks/dep-server/issues/196, where an already closed issue was re-opened

actions/issue/file/action.yml
```diff
@@ -57,15 +56,19 @@ runs:
           exit 1
         fi
 
-        release_issue=$(gh issue list --repo "${{ inputs.repo }}" --state all --label "${{ inputs.label }}" --json number --jq .[0].number)
+        issue_data=$(gh issue list --repo paketo-buildpacks/dep-server --state all --label workflow-failure --json title --json number --jq .[0])
```
fg-j, line 59:
```suggestion
        issue_data=$(gh issue list --repo "${{ inputs.repo }}" --state all --label "${{ inputs.label }}" --json title --json number --jq .[0])
```

fg-j: I think this jq search has a bit of a bug: Since we are only ever getting the 0th issue that has the specified label, we sometimes won't find the right issue that corresponds to the desired title. Imagine that `gh` query returns
```
[{ "number" : 2, 
  "title" : "Title B"
},
{ "number" : 2
  "title": "Title A"
}
]
```
and we're looking for the issue titled "Title A". We'll grab the 0th array element, but then fail the conditional on line 63 and skip ahead to filing a brand new issue.

fg-j: I think it'd be reasonable to filter the `gh` output based on title before grabbing the 0th element to achieve the desired outcome.

fg-j:
@sophiewigmore My mistake, I now see what you were going for with the title comparison. I agree it's logically equivalent! See my inline comment about a bug in the implementation.

fg-j (changes requested):

sophiewigmore:
@fg-j fixed to search for the specific issue title

fg-j (approved):

PR #520, opened 2022-08-02
Need to checkout branch automation/tools/go-get-update

joshuatcasey:
I'm seeing some issues running this workflow in practice, looks like the branch needs to be created first.

https://github.com/paketo-buildpacks/cpython/actions/workflows/go-get-update.yml

### DIFF
--- a/implementation/.github/workflows/go-get-update.yml
+++ b/implementation/.github/workflows/go-get-update.yml
@@ -18,6 +18,11 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v3
 
+      - name: Checkout Branch
+        uses: paketo-buildpacks/github-config/actions/pull-request/checkout-branch@main
+        with:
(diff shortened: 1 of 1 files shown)


ForestEckhardt (approved):

fg-j (commented):
~Do we also need this change in the language family workflow?~
Edit: Apparently the go-get-update workflow isn't part of the language family automations. @paketo-buildpacks/tooling-maintainers or @joshuatcasey can you remind me why not?

joshuatcasey:
I guess something got dropped in the conversation over here. https://github.com/paketo-buildpacks/github-config/pull/508#discussion_r929108402

I'd like to see this workflow working for implementation buildpacks before adding it in for language family buildpacks. I'll also try the `go get -u all` against several language familes before PRing that in.

fg-j (approved):

PR #530, opened 2022-08-11
add go.mod and go.sum to implementation patch list

sophiewigmore:
## Summary

Manually labeling tons of `Running 'go get -u all'`PRs  is getting tedious. It would be nice if these PRs auto-labeled and thus auto-merged. 

Edit: I have also modified the workflow to run once per week, rather than once per day.

Maintainer question - could it be dangerous to always auto-label `go.mod` changes as patches? Any package bumps that are breaking should hopefully cause linting or test failures which is why I think this is fine.

## Checklist

### DIFF
--- a/implementation/.github/.patch_files
+++ b/implementation/.github/.patch_files
@@ -25,3 +25,5 @@ scripts/.util/tools.sh
 scripts/.util/builders.sh
 scripts/integration.sh
 scripts/unit.sh
+go.mod
+go.sum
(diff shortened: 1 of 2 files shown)


sophiewigmore:
I would also be open to decreasing the frequency of this workflow to running once a week

robdimsdale:
> Maintainer question - could it be dangerous to always auto-label go.mod changes as patches? Any package bumps that are breaking should hopefully cause linting or test failures which is why I think this is fine.

I agree with labeling these bumps as `semver:patch`. We generally don't expose golang dependencies directly to our consumers, and we tend to integration-test everything (except maybe performance) so if there are any issues with dependencies that don't result in compilation failures, they'll probably result in lint failures or test failures.

fg-j (approved):

PR #531, opened 2022-08-11
Use max time rather than max attemps for exponential backoff.

robdimsdale:
## Summary

This PR changes the way we leverage exponential backoff to continue attempting the task until a duration is exceed (max time) rather than until a specific number of attempts have been made (max retries). This better aligns with what we care about - namely how long we have been trying to perform an action rather than how many times we have attempted it.

The change from 3 retries to 5 minutes also substantially increases how long we are willing to wait for a task to complete. In particular, we have observed that getting `503` while looking for USNs can take minutes, not seconds, to being working again. This will result in fewer failures getting USNs. Changing the create-release file upload is not necessary, but is done to reduce the cognitive overhead of having different limits on tasks with exponential backoff retries.

I also fixed the printing of how long we will wait until the next iteration as previously it was saying:

```
Retrying in 552.330144ms seconds
```

I had to make the timeout configurable to be able to integration test it, but I don't think it's particularly useful otherwise. If you have better suggestions as to how to make that timeout lower only for integration tests I'm all ears.

## Checklist

### DIFF
--- a/actions/release/create/entrypoint/main.go
+++ b/actions/release/create/entrypoint/main.go
@@ -26,12 +26,13 @@ type Release struct {
 
 func main() {
 	var config struct {
-		Endpoint string
-		Repo     string
-		Token    string
-		Release  Release
-		Draft    bool
-		Assets   string
+		Endpoint       string
+		Repo           string
+		Token          string
+		Release        Release
+		Draft          bool
+		Assets         string
+		RetryTimeLimit string
 	}
 
 	flag.StringVar(&config.Endpoint, "endpoint", "https://api.github.com", "Specifies endpoint for sending requests")
@@ -43,6 +44,7 @@ func main() {
 	flag.StringVar(&config.Release.Body, "body", "", "Contents of release body")
 	flag.BoolVar(&config.Draft, "draft", false, "Sets the release as a draft")
 	flag.StringVar(&config.Assets, "assets", "", "JSON-encoded assets metadata")
+	flag.StringVar(&config.RetryTimeLimit, "retry-time-limit", "5m", "How long to retry failures for")
 	flag.Parse()
 
 	if config.Repo == "" {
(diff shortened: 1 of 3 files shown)


sophiewigmore:
> Changing the create-release file upload is not necessary, but is done to reduce the cognitive overhead of having different limits on tasks with exponential backoff retries.
I don't think it would cause any cognitive overhead having different time limits. I'm fine to stick with using time rather than a set number of retries but I'd like to see the default set to a lower amount of time, even if it technically shouldn't make a difference

robdimsdale:
> I don't think it would cause any cognitive overhead having different time limits. I'm fine to stick with using time rather than a set number of retries but I'd like to see the default set to a lower amount of time, even if it technically shouldn't make a difference

Do you have a specific concern about the current maximum duration for retrying file upload during create release? I'm trying to understand the value in having it lower.

sophiewigmore:
I just don't really get the cognitive overhead concern here + in the off chance the release workflow does fail for a different reason than the upload, I'd rather it fails quickly after a minute or two. Maintainers have shepard releases out with this automation often, and I'd just rather have a faster feedback loop, as in, never hitting a case where I have to wait 5 minutes for something to retry over and over again

sophiewigmore:
This feels like splitting hairs though, since that case won't come up that often. I'd rather see it lowered for the reasons stated but whatever you think, I'll add my approval

sophiewigmore (dismissed):

robdimsdale:
That's fair. I see the value in a shorter feedback loop on failure for steps that are more likely to be triggered (and babysat) by a human. I'll drop the create-release upload max retry to one minute. I worry that significantly less than one minute might result in retrying only one (or even not at all) for larger files that fail towards the end of their upload.

sophiewigmore (approved):

PR #533, opened 2022-08-16
Use GITHUB_TOKEN to create release.

robdimsdale:
## Summary

When creating a release, we can use the provided `GITHUB_TOKEN` secret rather than a bot's token. We should use this token where possible, as it more secure due to being dynamically generated per-run and being much harder to leak.

I have validated that this works when creating releases in other workflows.

## Checklist

### DIFF
--- a/stack/.github/workflows/create-release.yml
+++ b/stack/.github/workflows/create-release.yml
@@ -533,7 +533,7 @@ jobs:
       uses: paketo-buildpacks/github-config/actions/release/create@main
       with:
         repo: ${{ github.repository }}
-        token: ${{ secrets.PAKETO_BOT_GITHUB_TOKEN }}
+        token: ${{ secrets.GITHUB_TOKEN }}
         tag_name: v${{ steps.tag.outputs.tag }}
         target_commitish: ${{ github.sha }}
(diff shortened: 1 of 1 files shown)


robdimsdale:
@ForestEckhardt yes, true, I didn't consider that. I think it's fine though, right? It's just swapping out one (unofficial) bot for another (official) bot.

ForestEckhardt:
Yeah like I said I think it is aesthetic and I am not concerned about it. Just wanted to call it out

fg-j (changes requested):
I don't think we can do this! If a release is created with this token, the Push Image workflows won't trigger.

ForestEckhardt:
> I don't think we can do this! If a release is created with this token, the Push Image workflows won't trigger.

Ooooooooo good catch. Do you know what is preventing that from happening?

fg-j:
@ForestEckhardt It's because Github Actions can't trigger other Github Actions. See https://github.com/googleapis/release-please/issues/922, another example of someone running into this.

robdimsdale:
Ahh, @fg-j thanks. I hadn't hooked up the push-image workflow in my other repos yet, so I hadn't seen that failure.

This is good to know! Thanks.

PR #542, opened 2022-08-30
Add workflows for library

joshuatcasey:
Add workflows for libraries.

The only use-case for this right now is https://github.com/paketo-buildpacks/libreload-packit, but it could be used for upcoming libraries such as `libdependency` (prototyped at https://github.com/joshuatcasey/libdependency) and we're really close to PRing these workflows into packit as well (https://github.com/paketo-buildpacks/packit/pull/385)!


library/scripts/integration.sh
```diff
@@ -0,0 +1,133 @@
+#!/usr/bin/env bash
```
ryanmoran, line 1: How does this integration test script fit into the library context?

library/scripts/.util/tools.json
```diff
@@ -0,0 +1,5 @@
+{
```
ryanmoran, line 1: We'll want to include changes to the workflow to update this file so it stays up-to-date: https://github.com/paketo-buildpacks/github-config/blob/f3bcea469f69d15417fbd75a600abbc6d9f4d91f/.github/workflows/update-tools.yml#L52-L99

library/.github/actions/release/notes/action.yml
```diff
@@ -0,0 +1,30 @@
+name: 'Create Buildpack Repo Release Notes'
```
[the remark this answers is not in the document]
joshuatcasey: That makes sense. Let's use GitHub magic.

library/.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,39 @@
...
+  push:
+    branches:
+    - main
+    - v*
+  pull_request:
+    branches:
+    - main
+    - v*
+  schedule:
+  - cron: '0 0 * * *' # Once a day at midnight
```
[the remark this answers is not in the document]
joshuatcasey: done (via rebase)

library/scripts/.util/tools.json
joshuatcasey: Hmm.. I hadn't seen this file before. Not sure what the action is to make sure it is updated.

joshuatcasey:
@ryanmoran @ForestEckhardt could you take a look again? I force-pushed to pull in latest from `main`.

ForestEckhardt (approved):
This all looks good to me. @ryanmoran is everything resolved on your end?

library/scripts/.util/tools.json
joshuatcasey: Oh I see. For now I'll actually remove `tools.json` since `jam` and `pack` (etc) aren't being used by `libdependency` and I'm not sure that they are needed for some general library integration testing. Will push that in a minute.

joshuatcasey: Resolving since library no longer has `tools.json`.

ryanmoran (approved):

ForestEckhardt (approved):

PR #563, opened 2022-09-16
Update Patch Files [Implementation, Language Family]

joshuatcasey:
Update Patch Files [Implementation, Language Family]

### DIFF
--- a/implementation/.github/.patch_files
+++ b/implementation/.github/.patch_files
@@ -4,13 +4,13 @@
 .github/dependabot.yml
 .github/labels.yml
 .github/workflows/approve-bot-pr.yml
-.github/workflows/auto-merge.yml
-.github/workflows/check-pr-labels.yml
 .github/workflows/codeql-analysis.yml
+.github/workflows/create-draft-release.yml
 .github/workflows/go-get-update.yml
 .github/workflows/label-pr.yml
 .github/workflows/lint-yaml.yml
 .github/workflows/lint.yml
+.github/workflows/push-buildpackage.yml
 .github/workflows/synchronize-labels.yml
 .github/workflows/test-pull-request.yml
 .github/workflows/update-dependencies.yml
@@ -22,8 +22,11 @@ README.md
 go.mod
 go.sum
 scripts/.util/builders.sh
+scripts/.util/git.sh
 scripts/.util/print.sh
 scripts/.util/tools.json
 scripts/.util/tools.sh
+scripts/build.sh
 scripts/integration.sh
+scripts/package.sh
 scripts/unit.sh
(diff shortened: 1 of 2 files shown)


ryanmoran:
I can't think of anything we should really be that concerned about.

sophiewigmore:
@joshuatcasey @ryanmoran context on why `build.sh` was omitted: https://github.com/paketo-buildpacks/github-config/pull/360#issuecomment-953902335

joshuatcasey:
> @joshuatcasey @ryanmoran context on why `build.sh` was omitted: [#360 (comment)](https://github.com/paketo-buildpacks/github-config/pull/360#issuecomment-953902335)
> 
> > With time and evidence these may be added, but for now they will not be auto-labeled as patch changes.

I recommend we add them - do we have any evidence not to?

sophiewigmore:
It rarely changes so there's not really any benefit to having it or not having it. I don't think we have any evidence either way because it hasn't changed. I'm fine with leaving it off unless we see a situation in which it would be helpful, but that's just me.

implementation/.github/.patch_files
```diff
@@ -19,11 +19,14 @@
...
 README.md
+go.mod
+go.sum
+scripts/.util/builders.sh
+scripts/.util/git.sh
 scripts/.util/print.sh
 scripts/.util/tools.json
 scripts/.util/tools.sh
-scripts/.util/builders.sh
+scripts/build.sh
```
fg-j, line 29:
```suggestion
```

joshuatcasey: will commit separately

joshuatcasey:
Updated to remove `scripts/build.sh` from the implementation patch files.

implementation/.github/.patch_files
```diff
@@ -4,13 +4,13 @@
 .github/dependabot.yml
 .github/labels.yml
 .github/workflows/approve-bot-pr.yml
-.github/workflows/auto-merge.yml
-.github/workflows/check-pr-labels.yml
```
sophiewigmore, line 8: why are these getting removed

sophiewigmore:
@joshuatcasey you removed `build.sh` but added `create-release`, `push-buildpackage`, and `package.sh` which were all originally omitted for the same initial reason as `build.sh`. I'm just a little confused - why removed `build.sh` but leave those? Is there different reasoning for them?

joshuatcasey:
> @joshuatcasey you removed `build.sh` but added `create-release`, `push-buildpackage`, and `package.sh` which were all originally omitted for the same initial reason as `build.sh`. I'm just a little confused - why removed `build.sh` but leave those? Is there different reasoning for them?

@sophiewigmore No different reasoning, just didn't explore the original issue to see why these were omitted. They have been removed now.

sophiewigmore (approved):

PR #572, opened 2022-09-30
Add exponential backoff failure handling to release asset downloading.

robdimsdale:
## Summary

This PR adds retry logic with exponential backoff to the download-asset action. To achieve this, this PR rewrites the action in Go.

A side benefit of writing this in Go is that it makes testing feasible, so this PR adds tests for the action.

## Use Cases

Roughly half of the monthly CI failures across the various stacks pipelines are issues downloading release assets that would likely be succesful upon retry.

Therefore adding retry logic should significantly reduce the number of flakes in the stacks pipelines.

It will also reduce the number of flakes in other pipelines that leverage this action, too, though we haven't performed any failure analysis on other pipelines.

## Checklist


fg-j:
Does the action need an `action.yml`? It seems like other golang + Dockerfile actions do (e.g. the create release action)

fg-j:
@robdimsdale I notice that you don't test that the exponential backoff actually occurs, and that you also don't test when writing the output artifact fails. Are these omissions intentional?

robdimsdale:
@fg-j yeah, i thought about both of those things.

### Exponential backoff testing
The exponential backoff happens during the error cases, but it's not tested for explicitly. We could assert this happens by validating that output contains "retrying" for an error case.

I'm happy to add this, I was somewhat on the fence about it.

### Handling errors when writing to the (created) file

I couldn't find a good way to test this with the test infrastructure we have (i.e. the integration test). If you have a good way to force a failure to write to a file after it was successfully created, please show me!

fg-j:
I'm ok with not testing these things! The code is quite simple and I see that you've got proper error handling around the failure-to-write case. Just wanted to make sure the omissions were intentional.

fg-j (approved):

PR #577, opened 2022-10-05
Add script to help with publishing draft releases.

robdimsdale:
## Summary

This PR adds a helper script to facilitate publishing draft releases of buildpacks. The intention is for this script to be run locally by a human, not automatically (e.g. not in CI).

An abbreviated example output is below:

```
❯ ./scripts/publish_draft_releases.sh --team 'python' --team 'ruby'

Getting teams in orgs [ paketo-buildpacks paketo-community ]
Getting repos for teams: [ paketo-buildpacks/python-maintainers paketo-buildpacks/ruby-maintainers ]

[ ... content abbreviated for clarity ... ]

Getting draft releases for paketo-buildpacks/python
↳ None found

Getting draft releases for paketo-buildpacks/python-start

Draft release found for paketo-buildpacks/python-start (v0.12.4)
Commits since previous published release (v0.12.3):
{
  "author": "Sophie Wigmore",
  "message": "Add new dependency workflow to .syncignore (#191)"
}
{
  "author": "dependabot[bot]",
  "message": "Bump github.com/paketo-buildpacks/packit/v2 from 2.4.2 to 2.6.1\n\nBumps [github.com/paketo-buildpacks/packit/v2](https://github.com/paketo-buildpacks/packit) from 2.4.2 to 2.6.1.\n- [Release notes](https://github.com/paketo-buildpacks/packit/releases)\n- [Commits](https://github.com/paketo-buildpacks/packit/compare/v2.4.2...v2.6.1)\n\n---\nupdated-dependencies:\n- dependency-name: github.com/paketo-buildpacks/packit/v2\n  dependency-type: direct:production\n  update-type: version-update:semver-minor\n...\n\nSigned-off-by: dependabot[bot] <support@github.com>"
}
{
  "author": "dependabot[bot]",
  "message": "Bump github.com/paketo-buildpacks/occam from 0.13.1 to 0.13.2\n\nBumps [github.com/paketo-buildpacks/occam](https://github.com/paketo-buildpacks/occam) from 0.13.1 to 0.13.2.\n- [Release notes](https://github.com/paketo-buildpacks/occam/releases)\n- [Commits](https://github.com/paketo-buildpacks/occam/compare/v0.13.1...v0.13.2)\n\n---\nupdated-dependencies:\n- dependency-name: github.com/paketo-buildpacks/occam\n  dependency-type: direct:production\n  update-type: version-update:semver-patch\n...\n\nSigned-off-by: dependabot[bot] <support@github.com>"
}
{
[...]

PR #594, opened 2022-10-21
Rewrite publish draft release script

robdimsdale:
## Summary

This PR rewrites the `publish_draft_releases.sh` script to use Go instead of Bash and `jq` filtering.

It is much easier to read and maintain and it is faster - especially when multiple teams are provided.

It still leverages the `gh` CLI for two reasons:
1. The CLI provides a convenient mechanism for authenticating: without it users would have to obtain and provide a token when invoking the script. Or we would have to make this an oauth app which is even more complicated.
2. It makes pagination easy via the `--paginate` flag.

We are still leveraging `jq` to turn the output of the `gh api --paginate` into a single valid array (via the `gh api --paginate | jq -s . | jq add` combination). Otherwise we would have to handle pagination ourselves and that is more complicated.

Finally, I kept the existing bash script and turned it into an entrypoint for the the golang script. This is to faciliate the transition from the bash script to the golang application. However, over time I'd like to remove the bash script and keep only the golang script.

There are no tests because there aren't any tests in the existing `bash` script 😄 

## Use Cases

As before in #577 - enable humans to easily publish draft releases of repositories they are maintainers of.

## Example output

Screen Shot 2022-10-21 at 4 49 38 PM

## Checklist


sophiewigmore (approved):

sophiewigmore:
works functionally the same and I find the Go easier to read 👍

PR #604, opened 2022-10-28
Replace deprecated GHA ::set-output syntax

### LINKED ISSUE #588: Replace deprecated GHA ::set-ouput syntax
Looks like we will need to update the way we output info from jobs in github actions: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

TL;DR - use this syntax instead:

```
echo "{name}={value}" >> $GITHUB_OUTPUT
```

e.g.
```
- name: Set color
  id: random-color-generator
  run: echo "SELECTED_COLOR=green" >> $GITHUB_OUTPUT
- name: Get color
  run: echo "The selected color is ${{ steps.random-color-generator.outputs.SELECTED_COLOR }}"
```

phil9909:
## Summary

Use the new way of producing outputs in a github action as the old way is deprecated and will be removed (31st May 2023).

See https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

If this PR is to big to review, I can break it down into smaller PRs. 

## Use Cases

Fixes #588

## Checklist

### DIFF
--- a/actions/release/reset-draft/entrypoint/main.go
+++ b/actions/release/reset-draft/entrypoint/main.go
@@ -93,7 +93,17 @@ func main() {
 		fail(fmt.Errorf("unexpected response from delete draft release request: %s", dump))
 	}
 
-	fmt.Printf("::set-output name=current_version::%s\n", releases[0].TagName)
+	outputFileName, ok := os.LookupEnv("GITHUB_OUTPUT")
+	if !ok {
+		fail(errors.New("GITHUB_OUTPUT is not set, see https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-an-output-parameter"))
+	}
+	file, err := os.OpenFile(outputFileName, os.O_APPEND|os.O_CREATE|os.O_WRONLY, 0666)
+	if err != nil {
+		fail(err)
+	}
+	defer file.Close()
+	fmt.Fprintf(file, "current_version=%s\n", releases[0].TagName)
+
 	fmt.Println("Success!")
 }
 
--- a/actions/release/reset-draft/entrypoint/main_test.go
+++ b/actions/release/reset-draft/entrypoint/main_test.go
@@ -7,7 +7,10 @@ import (
 	"net/http"
 	"net/http/httptest"
 	"net/http/httputil"
+	"os"
 	"os/exec"
+	"path/filepath"
+	"strings"
 	"testing"
 	"time"
 
@@ -34,6 +37,7 @@ func TestEntrypoint(t *testing.T) {
 
 			api      *httptest.Server
 			requests []*http.Request
+			tempDir  string
 		)
 
 		it.Before(func() {
@@ -110,9 +114,11 @@ func TestEntrypoint(t *testing.T) {
 					fmt.Fprintln(w, `{"error": "server-error"}`)
 
 				default:
-					t.Fatal(fmt.Sprintf("unknown request: %s", dump))
+					t.Fatalf("unknown request: %s", dump)
 				}
 			}))
+
+			tempDir = t.TempDir()
 		})
 
 		context("when a draft release does NOT exists", func() {
@@ -149,6 +155,10 @@ func TestEntrypoint(t *testing.T) {
 					"--repo", "some-org/draft-repo",
 					"--token", "some-github-token",
 				)
+				command.Env = []string{
+					fmt.Sprintf("GITHUB_OUTPUT=%s", filepath.Join(tempDir, "github-output")),
+					fmt.Sprintf("GITHUB_STATE=%s", filepath.Join(tempDir, "github-state")),
+				}
 
 				buffer := gbytes.NewBuffer()
 
@@ -166,8 +176,12 @@ func TestEntrypoint(t *testing.T) {
 				Expect(buffer).To(gbytes.Say(`Fetching latest releases`))
 				Expect(buffer).To(gbytes.Say(`  Repository: some-org/draft-repo`))
 				Expect(buffer).To(gbytes.Say(`Latest release is draft, deleting.`))
-				Expect(buffer).To(gbytes.Say(`::set-output name=current_version::some-version`))
 				Expect(buffer).To(gbytes.Say(`Success`))
+
+				data, err := os.ReadFile(filepath.Join(tempDir, "github-output"))
+				Expect(err).NotTo(HaveOccurred())
+				outputs := strings.Split(string(data), "\n")
+				Expect(outputs).To(ContainElements("current_version=some-version"))
 			})
 		})
 
--- a/actions/stack/diff-package-receipts/entrypoint/main.go
+++ b/actions/stack/diff-package-receipts/entrypoint/main.go
@@ -142,9 +142,18 @@ func main() {
 		)
 	}
 
-	fmt.Printf("::set-output name=added::%s\n", string(addedJSON))
-	fmt.Printf("::set-output name=removed::%s\n", string(removedJSON))
-	fmt.Printf("::set-output name=modified::%s\n", string(modifiedJSON))
+	outputFileName, ok := os.LookupEnv("GITHUB_OUTPUT")
+	if !ok {
+		log.Fatal("GITHUB_OUTPUT is not set, see https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-an-output-parameter")
+	}
+	file, err := os.OpenFile(outputFileName, os.O_APPEND|os.O_WRONLY, 0)
+	if err != nil {
+		log.Fatal(err)
+	}
+	defer file.Close()
+	fmt.Fprintf(file, "added=%s\n", string(addedJSON))
+	fmt.Fprintf(file, "removed=%s\n", string(removedJSON))
+	fmt.Fprintf(file, "modified=%s\n", string(modifiedJSON))
 }
 
 func parsePackagesFromFile(path string) (map[string]CycloneDXComponent, error) {
--- a/actions/stack/get-usns/entrypoint/main.go
+++ b/actions/stack/get-usns/entrypoint/main.go
@@ -126,7 +126,16 @@ func main() {
 	}
 
 	fmt.Println("Output: ", string(output))
-	fmt.Printf("::set-output name=usns::%s\n", string(output))
+	outputFileName, ok := os.LookupEnv("GITHUB_OUTPUT")
+	if !ok {
+		log.Fatal("GITHUB_OUTPUT is not set, see https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-an-output-parameter")
+	}
+	file, err := os.OpenFile(outputFileName, os.O_APPEND|os.O_WRONLY, 0)
+	if err != nil {
+		log.Fatal(err)
+	}
+	defer file.Close()
+	fmt.Fprintf(file, "usns=%s\n", string(output))
 
 	if config.Output != "" {
 		path, err := filepath.Abs(config.Output)
--- a/actions/stack/release-notes/entrypoint/main.go
+++ b/actions/stack/release-notes/entrypoint/main.go
@@ -2,13 +2,14 @@ package main
 
 import (
 	"bytes"
+	"crypto/rand"
 	_ "embed"
+	"encoding/hex"
 	"encoding/json"
 	"flag"
 	"fmt"
 	"log"
-	"regexp"
-	"strings"
+	"os"
 	"text/template"
 )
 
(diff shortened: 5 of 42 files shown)


builder/.github/workflows/create-release.yml
```diff
@@ -112,4 +112,4 @@ jobs:
         issue_body: |
           Create Release workflow [failed](https://github.com/${{github.repository}}/actions/runs/${{github.run_id}}).
         comment_body: |
-           Another failure occurred: https://github.com/${{github.repository}}/actions/runs/${{github.run_id}}
+          Another failure occurred: https://github.com/${{github.repository}}/actions/runs/${{github.run_id}}
```
robdimsdale, line 115: I assume this whitespace change is still valid yaml - is the indentation change a problem?

actions/release/create/entrypoint/main_test.go
```diff
@@ -133,6 +135,27 @@ func TestEntrypoint(t *testing.T) {
...
+			file, err = os.CreateTemp("", "state")
+			Expect(err).NotTo(HaveOccurred())
+			githubSaveStateFile = file.Name()
+			Expect(file.Close()).To(Succeed())
+		})
+
+		it.After(func() {
+			if githubSetOutputFile != "" {
+				Expect(os.Remove(githubSetOutputFile)).To(Succeed())
+				githubSetOutputFile = ""
```
robdimsdale, line 153: I think we can replace this with `os.RemoveAll` and drop the `if` condition. That's analogous to `rm -f` - it will still succeed if the file is not present.

actions/release/create/entrypoint/main_test.go
```diff
@@ -35,8 +35,10 @@ func TestEntrypoint(t *testing.T) {
 			Expect     = NewWithT(t).Expect
 			Eventually = NewWithT(t).Eventually
 
-			api      *httptest.Server
-			requests []*http.Request
+			api                 *httptest.Server
```
robdimsdale, line 38: I see that we changed the `main_test.go` for the `release/create` action, but I don't see any changes to the `main.go` file for the `release/create` action itself.

Am I missing something?

actions/release/reset-draft/entrypoint/main.go
```diff
@@ -93,7 +93,17 @@ func main() {
 		fail(fmt.Errorf("unexpected response from delete draft release request: %s", dump))
 	}
 
-	fmt.Printf("::set-output name=current_version::%s\n", releases[0].TagName)
+	outputFileName, ok := os.LookupEnv("GITHUB_OUTPUT")
+	if !ok {
+		fail(errors.New("GITHUB_OUT is not set, see https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-an-output-parameter"))
```
robdimsdale, line 98: Similar to above - I think this error message should read: `GITHUB_OUTPUT is not set`

robdimsdale: Also I'm probably fine not checking this `ok` condition - GitHub guarantees that this environment variable is set, and even if it's not set we will just fail on the next line, albeit with a less helpful error message.

I'm happy to leave this check in the code, but I'd be happy to see it removed either. I'm just pointing out the user experience is pretty much identical either way.

actions/release/reset-draft/entrypoint/main_test.go
```diff
@@ -110,9 +114,30 @@ func TestEntrypoint(t *testing.T) {
...
+			file, err = os.CreateTemp("", "state")
+			Expect(err).NotTo(HaveOccurred())
+			githubSaveStateFile = file.Name()
+			Expect(file.Close()).To(Succeed())
+		})
+
+		it.After(func() {
+			if githubSetOutputFile != "" {
+				Expect(os.Remove(githubSetOutputFile)).To(Succeed())
+				githubSetOutputFile = ""
```
robdimsdale, line 135: As above, can we use `os.RemoveAll` and drop the `if` condition?

robdimsdale: Also, you could use [`t.TempDir`](https://pkg.go.dev/testing#T.TempDir) with a `filepath.Join` - that will sidestep the requirement to clean up after each test.

actions/stack/diff-package-receipts/entrypoint/main.go
```diff
@@ -142,9 +142,18 @@ func main() {
 		)
 	}
 
-	fmt.Printf("::set-output name=added::%s\n", string(addedJSON))
-	fmt.Printf("::set-output name=removed::%s\n", string(removedJSON))
-	fmt.Printf("::set-output name=modified::%s\n", string(modifiedJSON))
+	outputFileName, ok := os.LookupEnv("GITHUB_OUTPUT")
+	if !ok {
+		log.Fatal("GITHUB_OUT is not set, see https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-an-output-parameter")
```
robdimsdale, line 147: Another instance where the error message should say: `GITHUB_OUTPUT`

actions/stack/get-usns/entrypoint/main.go
```diff
@@ -126,7 +126,16 @@ func main() {
 	}
 
 	fmt.Println("Output: ", string(output))
-	fmt.Printf("::set-output name=usns::%s\n", string(output))
+	outputFileName, ok := os.LookupEnv("GITHUB_OUTPUT")
+	if !ok {
+		log.Fatal("GITHUB_OUT is not set, see https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-an-output-parameter")
```
robdimsdale, line 131: Another instance where the error message should say: `GITHUB_OUTPUT`

actions/stack/release-notes/entrypoint/main.go
```diff
@@ -122,18 +123,26 @@ func main() {
 
 	fmt.Println(b.String())
 
-	fmt.Println(fmt.Sprintf("::set-output name=release_body::%s", escape(b.String())))
+	outputFileName, ok := os.LookupEnv("GITHUB_OUTPUT")
+	if !ok {
+		log.Fatalf("GITHUB_OUT is not set, see https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-an-output-parameter")
```
robdimsdale, line 128: Another instance where the error message should say: `GITHUB_OUTPUT`

actions/stack/release-notes/entrypoint/main.go
```diff
@@ -122,18 +123,26 @@ func main() {
...
-	newline := regexp.MustCompile(`\n`)
-	cReturn := regexp.MustCompile(`\r`)
-
-	result := strings.ReplaceAll(original, `%`, `%25`)
-	result = newline.ReplaceAllString(result, `%0A`)
-	result = cReturn.ReplaceAllString(result, `%0D`)
-
-	return result
+func generateDelimiter() string {
+	data := make([]byte, 16) // roughly the same entropy as uuid v4 used in https://github.com/actions/toolkit/blob/b36e70495fbee083eb20f600eafa9091d832577d/packages/core/src/file-command.ts#L28
```
robdimsdale, line 140: I'd also be happy to use the https://github.com/google/uuid package instead of creating our own.

I think it's slightly clearer to use the `uuid` package as it captures the intent of what we're doing (and is fewer steps so less likely to accidentally change something in the future).

But I think it's fine either way.

actions/tag/calculate-semver/entrypoint/main.go
```diff
@@ -236,3 +236,16 @@ func calculateNextSemver(previous semver.Version, largestChange int) semver.Vers
 	}
 	return semver.Version{}
 }
+
+func setOutput(tag string) {
```
robdimsdale, line 240: Can we rename this function to more clearly capture that it is only writing tags? Initially I thought it was writing any key/value pair, but now I see it is only writing tags.

Maybe call it: `writeTagOutput`?

actions/tag/calculate-semver/entrypoint/main_test.go
```diff
@@ -200,11 +204,39 @@ func TestEntrypoint(t *testing.T) {
...
+
+			file, err = os.CreateTemp("", "state")
+			Expect(err).NotTo(HaveOccurred())
+			githubSaveStateFile = file.Name()
+			Expect(file.Close()).To(Succeed())
 		})
 
+		it.After(func() {
+			if githubSetOutputFile != "" {
+				Expect(os.Remove(githubSetOutputFile)).To(Succeed())
```
robdimsdale, line 224: As above, does `os.RemoveAll` (or `t.TempDir()`) simplify this?

actions/tag/calculate-semver/entrypoint/main_test.go
```diff
@@ -200,11 +204,39 @@ func TestEntrypoint(t *testing.T) {
...
+				githubSetOutputFile = ""
+			}
+			if githubSaveStateFile != "" {
+				Expect(os.Remove(githubSaveStateFile)).To(Succeed())
+				githubSaveStateFile = ""
+			}
+		})
+
+		outputContains := func(elements ...string) {
```
robdimsdale, line 233: Have you tested what happens when this fails? Specifically, does the failure message get associated with the correct line?

I have vague memories of using helper functions like this resulting in incorrect line info, but that might have been specific to `ginkgo`, rather than `go test` + `spec` library.

robdimsdale (changes requested):
Thank you so much for submitting this!

I agree that it's long, but I don't think we need to split it into separate commits or PRs. All the changes are part of the same logical change; it's just a change that touches a lot of files.

There's a couple of typos (and maybe a missing file?) that I think we need to fix before we can merge this, and then a couple of potential refactors that won't change the functionality but might improve the maintainability of the code.

Other than that, I think this would be good to go. I'm not a tooling maintainer so obviously we need to wait for them to approve as well. Lots of folks are out this week at Kubecon North America, so it might be a few days before it can be merged.

Out of curiosity, did you do this all by hand or did you script any of it (e.g. with `sed`)?

builder/.github/workflows/create-release.yml
phil9909: This is actually unrelated to the change, but I couldn't resist to fix it anyway (notice the there are 3 more spaces in front of `Another` compared to `comment_body`, everywhere else "one level" is 2 spaces).

Still, it is unrelated, so I can remove it from this PR, if you prefer.

actions/release/create/entrypoint/main_test.go
phil9909: changed the wrong test file, will revert this part.

actions/tag/calculate-semver/entrypoint/main_test.go
phil9909: The line reported will be in the helper function. Actually, there is [`ExpectWithOffset`](https://onsi.github.io/gomega/#making-assertions-in-helper-functions) for this use-cases, but the offset seems to be ignored (maybe only works with `ginkgo`?)

Example output:

```
--- FAIL: TestEntrypoint (0.69s)
    --- FAIL: TestEntrypoint/calculate-semver (0.22s)
        --- FAIL: TestEntrypoint/calculate-semver/all_PRs_since_the_last_release_have_semver:patch/increments_the_patch_from_the_previous_version (0.01s)
            main_test.go:231: 
                Expected
                    <[]string | len:2, cap:2>: ["tag=1.2.4", ""]
                to contain elements
                    <[]string | len:1, cap:1>: ["tag=1.2.5"]
                the missing elements were
                    <[]string | len:1, cap:1>: ["tag=1.2.5"]
FAIL
FAIL	github.com/paketo-buildpacks/github-config/actions/tag/calculate-semver-tag/entrypoint	0.696s
FAIL
```

While the line is not to helpful, the test name is printed, which should be sufficient to find the relevant part, WDYT?
Alternatively, I could make the function return the output array and move the matcher call to to individual tests, if you prefer.

actions/release/reset-draft/entrypoint/main.go
phil9909: I don't mind either, anyone else with a strong opinion on this?

actions/stack/release-notes/entrypoint/main.go
phil9909: I don't mind either, anyone else with a strong opinion on this?

phil9909:
> Out of curiosity, did you do this all by hand or did you script any of it (e.g. with `sed`)?

Semi-automated I'd say. Obviously some parts required special care (e.g. the go tests and the multi-line stuff), but most of it could be solved by VS Code "search and replace" which supports regex and capture groups in the replacement. I guess `sed` would have worked the same.

Search: `echo "::set-output name=([^:]+)::(.+)"`
Replacement: `echo "$1=$2" >> "$GITHUB_OUTPUT"`

actions/tag/calculate-semver/entrypoint/main_test.go
robdimsdale: I think the `ExpectWithOffset` is what I was thinking of, and it doesn't surprise me that it's only required/only works when using Ginkgo.

I think this is fine - you can see which test is failing and what the expected/desired cases are.

builder/.github/workflows/create-release.yml
robdimsdale: yeah, I appreciate the cleanup! I'm happy to keep it in the PR.

I just wanted to make sure that the indentation doesn't matter (which I think it doesn't because it's effectively ignored in a multi-line string).

robdimsdale (changes requested):
These changes look good.

I think I only have one outstanding question - can we simplify the tests via `t.TempDir()` (or `os.RemoveAll()` if `t.TempDir()` doesn't work)?

Once we make that change, or can articulate why it doesn't work, then I'm happy to approve.

robdimsdale (dismissed):
Looks great - thanks for working through this!

ForestEckhardt (dismissed):

ForestEckhardt:
@paketo-buildpacks/tooling-maintainers Given the size of this change it seems to me that getting a majority of maintainer approval on this would make the most sense. I am going to hold off merging until I see more approvals.

phil9909:
@robdimsdale @ForestEckhardt My rebase has dismissed your approvals :disappointed: 

> I am would suggest holding off on merging until there are more approvals.

Fine for me

ForestEckhardt (approved):

ryanmoran (approved):

robdimsdale (approved):

phil9909:
> getting a majority of maintainer approval

Tooling maintainers are 4 people right? So with @ryanmoran and @ForestEckhardt approving we have a 50%, which is technically not a majority.

@fg-j @sophiewigmore one of you wanna have a look at this as well?

sophiewigmore (approved):

ForestEckhardt:
@phil9909 I think we have a super majority now. If we could get you to rebase this one more time so we could merge it that would be appreciated!

phil9909:
I rebased and made sure no new `::set-ouput` usages were introduced

ForestEckhardt (approved):

ForestEckhardt:
Thank you very much for this PR!

PR #631, opened 2022-11-17
Change method of upgrading implementation buildpack dependencies

fg-j:
## Summary

As seen in failures like [this .NET buildpack](https://github.com/paketo-buildpacks/dotnet-core-aspnet-runtime/actions/runs/3457967590/jobs/5771914923), sometimes `go get -u all` fails to upgrade buildpack transitive dependencies. The problems seem to come from an issue with module version resolution. I've had a hard time tracking down a clear explanation of the difference between `go get -u all` and `go get -u -t ./...`. I do know that the latter works as expected consistently, doesn't upgrade dependencies to RC versions, and works for buildpacks that have a `main.go` (aka implementation buildpacks). I've seen all of the above issues with `go get -u all`. If someone can offer a good explanation of the difference - or make a good case for _not_ making this change, I'd be interested to hear it. Otherwise, I think this will save maintainers a lot of CI toil.

See #508 for historical context on the initial implementation.

## Checklist

### DIFF
--- a/implementation/.github/workflows/go-get-update.yml
+++ b/implementation/.github/workflows/go-get-update.yml
@@ -25,14 +25,14 @@ jobs:
 
       - shell: bash
         run: |
-          go get -u all
+          go get -u -t ./...
           go mod tidy
 
       - name: Commit
         id: commit
         uses: paketo-buildpacks/github-config/actions/pull-request/create-commit@main
         with:
-          message: "Running 'go get -u all'"
+          message: "Running 'go get -u -t ./...'"
           pathspec: "."
           keyid: ${{ secrets.PAKETO_BOT_GPG_SIGNING_KEY_ID }}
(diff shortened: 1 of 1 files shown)


sophiewigmore:
@fg-j I do not have any answers to your questions about why that's a better option or an opinion on which we should go for- but I am definitely down to reduce toil related to this workflow. I think updating transitive dependencies is important but the amount I have to deal with the related workflow failures is a lot.

I am open to just trying this change out and seeing how it goes. It's a pretty low-stakes workflow, and we can always roll back the change if we encounter strange behaviour.

sophiewigmore (approved):

sophiewigmore:
@paketo-buildpacks/tooling-maintainers speak now if you have any other opinions!

robdimsdale (approved):

PR #665, opened 2023-01-20
issue file act: output node-id of created/mod issue

arjun024:
Global node ID[1] of the issue can be used in subsequent jobs to perform an action on the created/edited issue.
E.g. add the issue to a project board etc.

gh cli does not yet have node-id support. So use gh graphql queries.

1: https://docs.github.com/en/graphql/guides/using-global-node-ids

## Checklist

* [x] I have viewed, signed, and submitted the Contributor License Agreement.
* [ ] I have linked issue(s) that this PR should close using keywords or the Github UI (See [docs](https://docs.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue))
* [ ] I have added an integration test, if necessary.
* [x] I have reviewed the [styleguide](https://github.com/paketo-buildpacks/community/blob/main/STYLEGUIDE.md) for guidance on my code quality.
* [x] I'm happy with the commit history on this PR (I have rebased/squashed as needed).

PR #666, opened 2023-01-20
add action that adds an issue to project and set fields

arjun024:
This action adds the given issue to a github project board and (optionally) sets a custom field of type Single-Select to the given value.
(e.g. in the Paketo Workstreams board, adding an issue and setting its "Workstream" field) This action is designed to be chained with `actions/issue/file`.

Here's a typical example of usage:
```yaml
    - name: File Issue
      id: file-issue
      uses: paketo-buildpacks/github-config/actions/issue/file@main
      with:
        token: <token>
        repo: ${{ github.repository }}
        issue_title: "Some title"
        issue_body: |
          Some issue body

    - name: Add issue to project
      id: issue-to-proj
      uses: paketo-buildpacks/github-config/actions/issue/add-to-project@action-issue-add-to-proj
      with:
        token: <token>
        issue-node-id: ${{ steps.file-issue.outputs.node-id }}
        project-org: my-org
        project-num: 24
        field-name: Workstream
        option-name: Third Workstream
```

It can also be used with any workflow that triggers off an issue event. E.g.
```yaml
on:
  issues:
    types:
      - opened
```
with input `issue-node-id` made available to the action as `{{ github.event.issue.node_id }}`.

Github CLI does not currently allow interacting with the API of new-style projects. GraphQL api doc:
https://docs.github.com/en/issues/planning-and-tracking-with-projects/automating-your-project/using-the-api-to-manage-projects

Related: https://github.com/paketo-buildpacks/github-config/pull/665

## Checklist

### DIFF
--- a/actions/issue/add-to-project/entrypoint
+++ b/actions/issue/add-to-project/entrypoint
@@ -1,0 +1,153 @@
+#!/usr/bin/env bash
+set -euo pipefail
+shopt -s inherit_errexit
+
+# Github CLI does not currently allow interacting with the API of new-style projects.
+# Github GraphQL API guide: https://docs.github.com/en/issues/planning-and-tracking-with-projects/automating-your-project/using-the-api-to-manage-projects
+
+function main() {
+  declare issue_node_id project_org project_num field_name option_name
+  declare -i project_num
+
+  while [[ "${#}" != 0 ]]; do
+    case "${1}" in
+      --token)
+        token="${2}"
+        shift 2
+        ;;
+
+      --issue-node-id)
+        issue_node_id="${2}"
+        shift 2
+        ;;
+
+      --project-org)
+        project_org="${2}"
+        shift 2
+        ;;
+
+      --project-num)
+        project_num=${2}
+        shift 2
+        ;;
+
+      --field-name)
+        field_name="${2}"
+        shift 2
+        ;;
+
+      --option-name)
+        option_name="${2}"
+        shift 2
+        ;;
+
+      "")
+        # skip if the argument is empty
+        shift 1
+        ;;
+
+      *)
+        echo "Unknown argument" "$@"
+        exit 1
+    esac
+  done
+
+  echo "${token}" | gh auth login --with-token
+
+  project_id=
+  field_id=
+  option_id=
+  fill_ids_from_input
+
+  project_item_id=
+  add_issue_to_proj
+
+  if [ -n "${field_name}" ] && [ -n "${option_name}" ]; then
+    set_proj_item_field
+  else
+    echo "field-name or option-name not set. Skipping setting project item field"
+  fi
+
+  printf "project-item-id=%s\n" "${project_item_id}" >> "$GITHUB_OUTPUT"
+}
+
+function fill_ids_from_input() {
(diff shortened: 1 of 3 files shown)


actions/issue/add-to-project/entrypoint
```diff
@@ -0,0 +1,152 @@
+#!/usr/bin/env bash
+set -euo pipefail
```
robdimsdale, line 2: @arjun024 would you mind also adding:

```
shopt -s inherit_errexit
```

It will ensure that the script fails with error when a subshell (`$()`) exits with non-zero code.

arjun024: Done. Thanks!

actions/issue/add-to-project/entrypoint
```diff
@@ -0,0 +1,153 @@
...
+        shift 1
+        ;;
+
+      *)
+        echo "Unknown argument" "$@"
+        exit 1
+    esac
+  done
+
+  echo "${token}" | gh auth login --with-token
```
robdimsdale, line 55: FYI we have two established patterns for providing authentication to `gh auth`. This is one pattern, and the other is to set the environment variable `$GITHUB_TOKEN` in the action config.

There are pros and cons to both approaches, and I don't think you should change this script to the model. I think I'd rather have a holistic discussion about which pattern we prefer and change all the scripts at once to which pattern we opt for.

I'm just making you aware in case you didn't already know about the other option.

actions/issue/add-to-project/entrypoint
```diff
@@ -0,0 +1,153 @@
...
+
+  project_id=
```
robdimsdale, line 57: I have a slight preference for `local project_id` rather than `project_id=` (or `project_id=""`). I think it more clearly maps onto the intention that the variable hasn't been set, rather than it was set and it's empty.

If you agree, would you be willing to update the various variables in this script?

actions/issue/add-to-project/entrypoint
```diff
@@ -0,0 +1,153 @@
...
+  field_id=
+  option_id=
+  fill_ids_from_input
+
+  project_item_id=
```
robdimsdale, line 62: I'd have a slight preference for `project_item_id` to be defined by the output of the `add_issue_to_proj` function (i.e. `project_item_id="$(add_issue_to_project)"` - obviously changing that function to return the new ID.

I think it's easier to keep track of variables this way, vs having them set in a more global manner and overridden inside other functions.

If you agree, would you be willing to make that change also?

robdimsdale: Also this will probably require you to pass that variable as an argument to the `set_proj_item_field()` function. I think that's ok. I think it makes it more obvious what functions require and what they modify.

actions/issue/add-to-project/entrypoint
arjun024: Done. See refactored code

actions/issue/add-to-project/entrypoint
arjun024: Done. See refactored code

robdimsdale (approved):
Looks great - thank you!

PR #750, opened 2023-04-14
install curl-dev in tool update action

### LINKED ISSUE #749: Failure: Update Tools workflow
Update Tools workflow [failed](https://github.com/paketo-buildpacks/github-config/actions/runs/4695374779).

sophiewigmore:
## Summary

Resolves #749 
We recently started seeing this workflow fail on the `actions/tools/latest` action step for updating `jam` with `curl: (48) An unknown option was passed in to libcurl`. Installing `curl-dev` resolves this error, although I'm not entirely sure _why_ this issue cropped up in the first place.

## Checklist

* [X] I have linked issue(s) that this PR should close using keywords or the Github UI (See [docs](https://docs.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue))

### DIFF
--- a/actions/tools/latest/Dockerfile
+++ b/actions/tools/latest/Dockerfile
@@ -3,6 +3,7 @@ FROM alpine/git
 RUN apk add \
       bash \
       curl \
+      curl-dev \
       jq \
     && rm -rf /var/cache/apk/*
 


robdimsdale:
@sophiewigmore can you also try with `libcurl` instead of `curl-dev`?

This [forum post](https://forums.freebsd.org/threads/curl-unknown-option-was-passed-to-libcurl.87969/) suggests that option, which seems better to me than installing `curl-dev`, because `curl-dev` provides the development headers  to link against `libcurl` which we probably don't need.

As far as why it changed, perhaps `apk add curl` is installing the newly-released `curl` v8 but isn't transitively pulling in `libcurl` v8 also? I dunno. Honestly it sounds like a problem with `apk`'s distribution but I don't necessarily want to wait for them to fix it.

sophiewigmore:
@robdimsdale I get a similar failure when I use `libcurl`:
`curl: (48) An unknown option was passed in to libcurl`

robdimsdale:
That is frustrating. It seems to me that `apk` is doing something inconsistent when installing `curl`.

Happy to merge this now, but maybe let's also file an issue to explore removing this library after waiting a while (e.g. a month) to see if `apk` has fixed whatever might be going wrong.

Separately, this is a +1 for pre-compiling our images - we would be able to pin our builds to a previous version of the image while we dig into this / wait for `apk` to resolve itself. However, we don't see this type of issue often so I don't think it's a particularly strong signal.

robdimsdale (approved):

robdimsdale:
Yes, i suppose that's fair.

PR #760, opened 2023-05-11
Use os/architecture specific tools, dealing with varying tool artifact naming conventions

### LINKED ISSUE #756: Use `arm64` based tools when building stacks on `arm64` architecture
Currently, the tools that are used by various dependent stack projects assume `amd64` architecture even on `arm64` machines. These tools don't work on all `arm64` machines so we should use architecture specific versions of these tools.

## Possible Solution
All the tools scripts (`tools.sh`) that are used by stack, builder, and buildpack projects can detect when they are running on an `arm64` architecture using the `uname -m` command. If running on `arm64`, the tools these scripts download/install should then be the `arm64` versions since the `arm64` versions of these tools are now available. `jam` was the last holdout, but since [`2.1.1`](https://github.com/paketo-buildpacks/jam/releases/tag/v2.1.1) version that is not the case.

In addition, the stack create script (`stack/scripts/create.sh`) can use the same architecture detection mechanism and use architecture specific `jam` command passing in an architecture specific TOML filename. This is necessary since the TOML file contains architecture specific values. When creating the stack on `arm64`, we can possible rely on a file named `stack-arm64.toml` to keep it separate from `stack.toml` which is used to build `amd64` architecture images for backwards compatibility.

## Possible Side Effects
The stack project will now have to maintain multiple TOML files, one for each architecture. There would be bunch of common content between these TOML files but there would be enough differences to warrant a separate file for each architecture. For example, the Ubuntu Jammy APT repositories are completely different for different architectures. In some cases, the TOML files might need to specify architecture specific APT packages as well.

Another side effect of this change would be that stack artifacts (OCI image archives) produced will always match the architecture they were built on due to the auto-detection of architecture in the *create* script. This means that simply running `scripts/create.sh` on a dependent project (like on [`paketo-buildpacks/jammy-tiny-stack`](https://github.com/paketo-buildpacks/jammy-tiny-stack)) on an `arm64` machine, the resulting image archives would always produce `arm64` versions.

## Motivation
The `amd64` version tools probably work OK on Apple M1 Macs due to Rosetta2 translation layer but won't work on other `arm64` machines. All these changes should be backwards compatible and won't effect building `amd64` stack image archives on an `amd64` machine but would be helpful for the brave amongst us trying to build `arm64` artifacts on `arm64` machines.

**I can submit a PR to get us started on this**

maliksalman:
## Summary
Currently, the tools that are used by various dependent projects assume `amd64` architecture even on `arm64` machines. These tools don't work on all `arm64` machines so we should use architecture specific versions of these tools when building stacks and buildpacks. Resolves #756

## Use Cases
The automation to build stacks and buildpacks should use architecture specific tools when running on `arm64` and `amd64` machines.

## Checklist

### DIFF
--- a/builder/scripts/.util/tools.sh
+++ b/builder/scripts/.util/tools.sh
@@ -6,6 +6,42 @@ set -o pipefail
 # shellcheck source=SCRIPTDIR/print.sh
 source "$(dirname "${BASH_SOURCE[0]}")/print.sh"
 
+function util::tools::os() {
+  case "$(uname)" in
+    "Darwin")
+      echo "${1:-darwin}"
+      ;;
+
+    "Linux")
+      echo "linux"
+      ;;
+
+    *)
+      util::print::error "Unknown OS \"$(uname)\""
+      exit 1
+  esac
+}
+
+function util::tools::arch() {
+  case "$(uname -m)" in
+    arm64|aarch64)
+      echo "arm64"
+      ;;
+
+    amd64|x86_64)
+      if [[ "${1:-}" == "--blank-amd64" ]]; then
+        echo ""
+      else
+        echo "amd64"
+      fi
+      ;;
+
+    *)
+      util::print::error "Unknown Architecture \"$(uname -m)\""
+      exit 1
+  esac
+}
+
 function util::tools::path::export() {
   local dir
   dir="${1}"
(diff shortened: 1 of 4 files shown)


robdimsdale (approved):

robdimsdale (approved):
Thanks for submitting this! I haven't tested it but it looks fine.

PR #779, opened 2023-07-11
action/find-asset: slice off first char only if it's a "v"

arjun024:
Don't slice off the first character from every tag. It's true the all current buildpack/stack repos use tags of format "vX.Y.Z" but this change will make it usable for repos that just use "X.Y.Z"

### DIFF
--- a/actions/release/find-asset/entrypoint
+++ b/actions/release/find-asset/entrypoint
@@ -62,7 +62,7 @@ function find_asset() {
     | jq -r \
     --arg pattern "${pattern}" \
     --argjson depth "${depth}" \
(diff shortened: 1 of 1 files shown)


arjun024:
[To maintainers] If merging, please also update the pre-built image for the action

robdimsdale (approved):

sophiewigmore:
@robdimsdale were you able to update the pre-built image?

sophiewigmore:
Thank you! That was the bit I was unsure about, aka why I hadn't reviewed yet :)

robdimsdale:
Ok, I just pushed the updated image for this action.

This was way too hard, so I'm going to try to automate it now while it's still fresh in my mind.

PR #825, opened 2023-10-05
Separate cron schedules for recurring workflows

robdimsdale:
## Summary

This PR separates the cron schedules for the various workflows. They were generated with psuedo-random number generators.

In most cases the overal frequency is unchanged, but in some cases (like stacks) the frequency is reduced substantially.

## Use Cases

We are starting to see issues with hitting GitHub API rate limits as well as workflow runs having infrastructure issues, and this is a quick and easy step to reduce the number of workflows running at the same time.

## Checklist

### DIFF
--- a/.github/workflows/update-tools.yml
+++ b/.github/workflows/update-tools.yml
@@ -3,7 +3,7 @@ name: Update Tools
 
 on:
   schedule:
-  - cron: '0 0 * * *'
+  - cron: '42 19 * * *' # daily at 19:42 UTC
   workflow_dispatch: {}
 
 concurrency: tools_update
--- a/stack/.github/workflows/create-release.yml
+++ b/stack/.github/workflows/create-release.yml
@@ -2,7 +2,7 @@ name: Create Release
 
 on:
   schedule:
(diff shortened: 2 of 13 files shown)


.github/workflows/update-tools.yml
```diff
@@ -3,7 +3,7 @@ name: Update Tools
 
 on:
   schedule:
-  - cron: '0 0 * * *'
+  - cron: '42 19 * * *' # daily at 19:42 UTC
```
sophiewigmore, line 6: there's a YAML linting error here and on a few other files

stack/.github/workflows/create-release.yml
```diff
@@ -2,7 +2,7 @@ name: Create Release
 
 on:
   schedule:
-  - cron: '*/30 * * * *'  # every 30 minutes
+  - cron: '27 2,14 * * *' # daily at 02:27 and 14:27 UTC
```
sophiewigmore, line 5: This is the only frequency change I'm wondering about, but down to try it and we can always adjust it up to be more frequent if needed

sophiewigmore (approved):
approved pending linting YAML fixes

PR #828, opened 2023-10-16
Improve debug output for increment tag action.

robdimsdale:
## Summary

This PR improves the debug output for the "increment tag" action.

## Use Cases

We recently ran into an issue where the output tag was not what we expected (see [this failing build](https://github.com/paketo-buildpacks/jammy-base-stack/actions/runs/6536023435/job/17746971933)). This PR improves the debug output so it is easier to reason about which logic path is being taken.

As a side note, I switched to `git tag --merged` (from `git rev-list --tags --max-count=1`) because it allows us to get a list of tags, not just the most recent one. This has the side effect of only listing tags that are in the current branch. This is probably what we want, as we should probably not include tags that aren't on the current branch, but I wanted to call it out explicitly.

## Checklist

### DIFF
--- a/actions/tag/increment-tag/entrypoint
+++ b/actions/tag/increment-tag/entrypoint
@@ -30,6 +30,9 @@ main() {
 
   git config --global --add safe.directory "${GITHUB_WORKSPACE}"
 
+  printf "current_version=%s\n" "${current_version}" >> "$GITHUB_OUTPUT"
+  printf "allow_head_tagged=%s\n" "${allow_head_tagged}" >> "$GITHUB_OUTPUT"
+
   if [ "${allow_head_tagged}" = "false" ]; then
     if git describe --exact-match --tags HEAD > /dev/null 2>&1; then
       echo "error: HEAD has already been tagged"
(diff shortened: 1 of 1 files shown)


actions/tag/increment-tag/entrypoint
```diff
@@ -38,14 +41,19 @@ main() {
...
-      echo "There are no previous tags"
+      printf "There are no previous tags - using v0.0.0 as the tag"
       previous="v0.0.0"
     else
       # List all tags of HEAD, sorted in ascending semver order; use the last one
-      previous="$(git tag --sort "version:refname" --points-at "$(git rev-list --tags --max-count=1)" | tail -n 1)"
+      all_previous="$(git tag --sort "version:refname" --merged)"
+      printf "Last 10 previous tags: %s\n" "$(tail -n 10 <<< "${previous}")"
+
+      previous="$(tail -n 1 <<< "${all_previous}")"
```
sophiewigmore, line 55: Not sure I get the benefit of doing this - why do we care about getting the last 10? Or how is it helpful for debugging?

robdimsdale: yeah it was just helpful for debugging because the most recent tag wasn't quite enough info to figure out why it was picking the wrong tag.

sophiewigmore: update - chatted with Rob about this - printing out 10 tags to help understand any weirdness in repo version history (as would have been useful in the failed stack build linked in the PR description)

sophiewigmore (approved):

PR #836, opened 2023-10-25
Remove `--merged` flag

ForestEckhardt:
- Added additional debug printing

### DIFF
--- a/actions/tag/increment-tag/entrypoint
+++ b/actions/tag/increment-tag/entrypoint
@@ -49,7 +49,7 @@ main() {
       previous="v0.0.0"
     else
       # List all tags of HEAD, sorted in ascending semver order; use the last one
-      all_previous="$(git tag --sort "version:refname" --merged)"
+      all_previous="$(git tag --sort "version:refname")"
       printf "Last 10 previous tags: %s\n" "$(tail -n 10 <<< "${all_previous}")"
 
       previous="$(tail -n 1 <<< "${all_previous}")"
@@ -59,5 +59,6 @@ main() {
   fi
 
   echo "tag=${tag#v}" >> "$GITHUB_OUTPUT"
+  printf "New tag: %s\n" "${tag}"
 }
 main "${@:-}"


sophiewigmore (approved):

robdimsdale:
So the point of `--merged` is that it ignores tags that aren't on main (assuming the commit we're running against is on main, which it should be if we're making a release). This protects against the case @sophiewigmore and I saw where a release workflow was run off a branch and a tag was created on a side branch. Therefore this tag never existed on `main`.

Once the tag was made on a side branch, when the workflow ran without `--merged`, it would always try to recreate the same tag and just pile up draft releases that wouldn't be publishable, as there's already a release with that tag. When it ran with `--merged`, it correctly identified that the tag was in appropriate and skipped over it, making the next tag correctly.

Switching to the current problem: I am very suspicious of GHA's environment here. It's not exiting early with `set -u` and we get different results running the same command locally in a docker container vs in GHA docker container

After re-running the workflow with the tmate session to poke around a bit, it appears that the problem is the way that the repo is checked out - it's essentially a bunch of shallow clones - one per tag:

```
runner@fv-az419-569:~/work/tanzu-dotnet-core/tanzu-dotnet-core$ git log --graph --abbrev-commit --decorate --pretty=oneline --all | head -n10
* 82d7d8a (grafted, HEAD -> ForestEckhardt-patch-1, origin/ForestEckhardt-patch-1) Remove functional parts of workflow
* eafe575 (grafted, tag: v2.8.2) Updates buildpacks in buildpack.toml (#663)
* 42710cc (grafted, tag: v2.8.1) Updating buildpacks in buildpack.toml
* 0f75a09 (grafted, tag: v2.8.0) Updating github-config
* 640932d (grafted, tag: v2.7.2) Updates buildpacks in buildpack.toml (#629)
* 1f26903 (grafted, tag: v2.7.1) Updates buildpacks in buildpack.toml (#624)
* fbd2207 (grafted, tag: v2.7.0) Adds new-relic buildpack to order group (#615)
* 5759aab (grafted, tag: v2.6.2) Updating github-config
* 0d1e7a0 (grafted, tag: v2.6.1) Updating github-config
* fcad945 (grafted, tag: v2.6.0) Bump github.com/paketo-buildpacks/packit/v2 from 2.9.1 to 2.10.0
```

This means the --merged flag won't work because there's no commit history linking the `HEAD` to the tags. I can't see how this --merged ever worked, because we've always been checking out shallow clones via the [checkout action](https://github.com/actions/checkout).

At this point our options are either:
1. Remove the `--merged` flag (this PR) and trust that humans won't accidentally make a tag of a branch. This is risky, because we've done it before and might do it again. But it takes up less space on disk and less time to do a shallow clone.
2. Checkout the full history (or some reasonable amount of history that will likely encompass the previous tag). This will take a bit longer and use up a bit more disk space, but protects against accidental rogue commits. It also more closely resembles what we do locally.

PR #854, opened 2023-11-09
Reset-draft action: add optional version flag

sophiewigmore:
## Summary

Adds optional `--version` flag to the reset draft release action. 
If a tag version is provided, the action will reset a draft release matching the version and no-op exit if no match is found
If no tag version is provided, the action will proceed as usual, deleting the latest draft release if one exists 

## Checklist

* [X] I have added an integration test, if necessary.

### DIFF
--- a/actions/release/reset-draft/entrypoint/main.go
+++ b/actions/release/reset-draft/entrypoint/main.go
@@ -15,11 +15,13 @@ func main() {
 		Endpoint string
 		Repo     string
 		Token    string
+		Version  string
 	}
 
 	flag.StringVar(&config.Endpoint, "endpoint", "https://api.github.com", "Specifies endpoint for sending requests")
 	flag.StringVar(&config.Repo, "repo", "", "Specifies repo for sending requests")
 	flag.StringVar(&config.Token, "token", "", "Github Authorization Token")
+	flag.StringVar(&config.Version, "version", "", "Optional specific release version to reset")
 	flag.Parse()
 
 	if config.Repo == "" {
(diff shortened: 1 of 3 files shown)


actions/release/reset-draft/entrypoint/main.go
```diff
@@ -69,14 +71,35 @@ func main() {
 		return
 	}
 
+	releaseToDelete := releases[0]
```
robdimsdale, line 74: Should we also remove the section below:

```

	if !releases[0].Draft {
		fmt.Println("Latest release is published, exiting.")
		return
	}
```

I think this logic is covered by both branches in the `for` loop you added.

actions/release/reset-draft/entrypoint/main_test.go
```diff
@@ -364,6 +448,9 @@ func TestEntrypoint(t *testing.T) {
 					Expect(buffer).To(gbytes.Say(`{"error": "server-error"}`))
 				})
 			})
+			context.Focus("when a specific version is passed in", func() {
```
robdimsdale, line 451: I think we should remove the `Focus` and it seems like we should either remove this empty context or add some tests to it.

sophiewigmore: yes absolutely

actions/release/reset-draft/entrypoint/main.go
sophiewigmore: Unfortunately, the `for` loop case I added only applies to when we pass in `config.Version`. In the case where no version is passed in we want to perform the `releases[0].Draft` check still. Maybe it would be clearer if I added that case as an `else if`?

robdimsdale: Ah, gotcha. I think this logic makes sense. It might be slightly easier to read as follows:

```
var releaseToDelete github.Release // or whatever the type actually is

if config.Version == "" {
  releaseToDelete = releases[0]
} else {
  // existing for loop that eventually sets releaseToDelete = r
}

fmt.Printf("found draft with version: '%s', deleting", releaseToDelete.Version)
```

But I don't feel strongly either way.

robdimsdale (approved):

actions/release/reset-draft/entrypoint/main.go
sophiewigmore: Fixed per your suggestion - I prefer that!

robdimsdale (approved):

PR #923, opened 2024-05-24
Execute unit tests several times in to prevent flakes

c0d1ngm0nk3y:
## Summary

To prevent flakes in the unit tests, we can just run them 5 times in PR validation. The run time should't be a issue and this would make it much harder for flakes to sneak in.

## Use Cases

In `go` flakes in the tests can happen quite easy since looping over a dict has not a stable order.

Motivation: https://github.com/paketo-buildpacks/npm-install/pull/689

## Checklist


sophiewigmore:
🤔 I'm not against this, but I'd much rather just fix the flaky test to be deterministic than add a retry.

c0d1ngm0nk3y:
> 🤔 I'm not against this, but I'd much rather just fix the flaky test to be deterministic than add a retry.

Oh no, this is not adding a retry. That would be giving up :) It executes the tests several times in different order and they have to be successful all the time. Imho, the only reason against such a check could be the runtime which shouldn't be a problem when we talk about unit tests for a pull requests.

That should make it much harder for flakes to come into the code base since they can be really annoying.

c0d1ngm0nk3y:
@sophiewigmore Do you see any downside?

robdimsdale:
> So is this something that you have seen happening or is it preventative?

There are flakes somewhere in the nodejs tests (I forget exactly which buildpack).

Personally, I feel ambivalent about this:

One the one hand, I'd love to make it harder for flakes to creep in. As someone who has had to re-trigger nodejs builds that are failing for flaky unit tests, i wish that never crept into the codebase to start with.

On the other hand, if we make this change and there are already flakes, it will essentially make it impossible to get a green build until we fix the flakes. Which we should do, of course, but we may not have the bandwidth to do that in the moment. Imagine there's a critical CVE in nodejs today - we don't have the bandwidth to dig into nodejs failures right now (we're a little short on maintainers) so we would probably end up changing the test count param back to `1` and manually retrying until it succeeded. Or worse - we would delete the flaky test rather than fixing it to ensure that we can ship the fix in a timely manner.

There's also the secondary cost that running the tests multiple times takes N times longer. But the buildpack unit tests are fairly quick so I don't see this as an issue.

On balance, I think this is a change that I support, though I probably wouldn't have made it myself.

robdimsdale (approved):

implementation/scripts/unit.sh
```diff
@@ -6,6 +6,8 @@ set -o pipefail
 readonly PROGDIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 readonly BUILDPACKDIR="$(cd "${PROGDIR}/.." && pwd)"
 
+GO_TEST_PARAMS="${GO_TEST_PARAMS:=''}"
```
robdimsdale, line 9: FWIW this can be slightly simplified to: `"${GO_TEST_PARAMS:-}"

c0d1ngm0nk3y:
> On the other hand, if we make this change and there are already flakes, it will essentially make it impossible to get a green build until we fix the flakes.

Then the fix gets out of the sudden priority which is a good think imho :)


> There's also the secondary cost that running the tests multiple times takes N times longer. But the buildpack unit tests are fairly quick so I don't see this as an issue.

I don't see a problem either and if unit tests are so slow that you cannot execute them several times, something should be changed anyway.

c0d1ngm0nk3y:
> @c0d1ngm0nk3y So is this something that you have seen happening or is it preventative?

In `npm-install` there was a flake for months and it seemed that nobody bothered to much. It was a bit annoying to fix it. So I thought this might be a cheap way to prevent flakes and also give some priority to fixing them.

c0d1ngm0nk3y:
If no one has objections, how to get this merged?

robdimsdale (approved):

robdimsdale:
@c0d1ngm0nk3y I've approved it and yet it to auto merge. But I can't rebase it for you. Once you rebase this PR it should be auto-merged.

PR #935, opened 2024-06-05
Make auto-label semver action fail if there are no patchfiles

ForestEckhardt:
The function of this task is to check if it can auto-label anything with a semver label and return the label if it can. It is run after a check to see if there are any semver labels already set and if there are not this action will run. I think that it would then follow logically that this action should fail if there are no labels to add meaning the whole check task will fail.

## Checklist

### DIFF
--- a/actions/pull-request/auto-semver-label/check-files-changed.sh
+++ b/actions/pull-request/auto-semver-label/check-files-changed.sh
@@ -77,7 +77,7 @@ function main() {
     if [ "${safe}" -eq "0" ]; then
       echo "Files changed that aren't on the patch allowlist"
       echo "label=" >> "$GITHUB_OUTPUT"
-      exit 0
+      exit 1
     fi
   done < <(gh api /repos/"${repo}"/pulls/"${number}"/files --jq '.[].filename')
 


robdimsdale (approved):

sophiewigmore:
@ForestEckhardt / @robdimsdale - sorry I thought I was good with this change - but why would we want the workflow to fail noisily with an error when we simply encounter a file we can't add a patch label to? That doesn't seem like an error to me, it seems like intended behaviour

Can you explain why having a failure is preferable?

robdimsdale:
For posterity:

Currently the way we use this action is as follows (edited for brevity):

```
jobs:
  autolabel:
    steps:
    - name: Check Minimal Semver Labels
      uses: mheap/github-action-required-labels@v3
      with:
        count: 1
        labels: semver:major, semver:minor, semver:patch
        mode: exactly

    - name: Auto-label Semver
      if: ${{ failure() }}
      uses: paketo-buildpacks/github-config/actions/pull-request/auto-semver-label@main
```

e.g. [this workflow](https://github.com/paketo-buildpacks/github-config/blob/89d04c2fd0e1f3f3188a4c26be2a69f2b17dc581/language-family/.github/workflows/label-pr.yml#L15C1-L33C61)

So the logic here is:

* check for existing labels
  * if existing labels - exit with success
  * if no existing labels - exit with failure
* If the job fails, try to add a label. Don't worry about whether this succeeds or not

This successfully adds labels (where it can) if there weren't any, but in this case the job is always reported as failing. This is very noisy and misleading.

I think it's better to have a structure that looks more like the following:

```
jobs:
  autolabel:
    steps:
    - name: Check Minimal Semver Labels
      id: check
      uses: mheap/github-action-required-labels@v5
      with:
        count: 1
        labels: semver:major, semver:minor, semver:patch
        mode: exactly
        exit_type: success

    - name: Auto-label Semver
      if: ${{ steps.check.outputs.status == 'failure' }}
      uses: paketo-buildpacks/github-config/actions/pull-request/auto-semver-label@main
```

This structure has the following logic:

* check for existing labels
  * if existing labels - exit with success
  * if no existing labels - record that fact and continue to the next step
* If there were no labels, try to add a label. **Fail the job if we can't add labels**

However, that last point in bold requires a change to the action, which is this PR

sophiewigmore:
@ForestEckhardt @robdimsdale can you explain a bit more about why we want this step to noisily fail?

sophiewigmore:
Ok still wondering why we need to fail the job if we can't add labels? Won't that happen a lot, since not every file is in the patch file list?

ForestEckhardt:
Yes it will fail but only if there is not an existing semver label and it could not apply one which is the eact scenario it should fail in because that means that the PR needs to be manually labeled. If it does not fail in this scenario the PR will pass all checks without a valid smver label attached to it.

sophiewigmore:
Ah yea the bit about not wanting the PR checks to pass is what I wasn't considering

sophiewigmore (approved):

PR #937, opened 2024-06-07
Body filepath

pacostas:
## Summary

With this PR, we extend the functionality of the create release action by passing through the newly added option `body-filepath`, a string that points to the file that has the release notes instead of passing/storing the release notes on the `body` argument. Specifically, the current functionality as is, allows to pass the body of the release notes through the `body` argument of the action. We extend this functionality by introducing a new argument called `body-filepath` which points to a file to read from the release notes. This is useful in case we want to ensure that we don't reach the shell arguments limit which results in the `too many arguments` error.  The `body-filepath` extra argument is optional and has a priority only when the `body` argument is empty and the `body-filepath` is not.

* We already use this action in our stack workflow https://github.com/paketo-community/ubi-base-stack/blob/main/actions/release/create/entrypoint/main.go 

* It does not change the default behaviour

## Use Cases

With this PR we ensure that no error will occur in case of a newly added stack produces too many changes, e.g., on the added packages of the build and run image.

## Checklist


pacostas:
@ForestEckhardt Can you take a look whenever you have time?

ForestEckhardt (approved):

PR #950, opened 2024-06-18
Stack - Select on which registries to push stack images

pacostas:
## Summary

This PR adds an extra option, which specifies on which registries the stack should push. Specifically, if on the root directory of the implementation, there is a file called `registries.json` with the below content
```
{
  "dockerhub": true,
  "GCR": false
}
``` 
it means that the action will push the stack images to `dockerhub` but NOT to `GCR`. 

In case the `registries.json` file does not exist, it fallbacks to the current behavior which is to push on both registries.

In case where an attribute is NOT specified, then the default value is `true`.

## Use Cases

This is useful for projects being under paketo-community organization, where GCR is not available.

## Checklist

### DIFF
--- a/stack/.github/workflows/push-image.yml
+++ b/stack/.github/workflows/push-image.yml
@@ -103,13 +103,7 @@ jobs:
       run: |
 
         if [ "${{ needs.preparation.outputs.push_to_dockerhub }}" == "true" ]; then
-        echo "${DOCKERHUB_PASSWORD}" | sudo skopeo login --username "${DOCKERHUB_USERNAME}" --password-stdin index.docker.io
-        echo "${DOCKERHUB_PASSWORD}" | sudo skopeo login --username "${DOCKERHUB_USERNAME}" --password-stdin index.docker.io
-        echo "${GCR_PASSWORD}" | sudo skopeo login --username "${GCR_USERNAME}" --password-stdin gcr.io
-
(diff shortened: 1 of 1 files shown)


pacostas:
@mhdawson @ForestEckhardt

stack/.github/workflows/push-image.yml
```diff
@@ -74,15 +101,24 @@ jobs:
         GCR_PASSWORD: ${{ secrets.GCR_PUSH_BOT_JSON_KEY }}
         GCR_PROJECT: "${{ github.repository_owner }}"
       run: |
+
+        if [ "${{ needs.preparation.outputs.push_to_dockerhub }}" == "true" ]; then
+        echo "${DOCKERHUB_PASSWORD}" | sudo skopeo login --username "${DOCKERHUB_USERNAME}" --password-stdin index.docker.io
```
robdimsdale, line 106: Looks like this line is here multiple times. Am I missing something?

robdimsdale (commented):
This looks good, with one minor question (inline) about potentially duplicated lines.

stack/.github/workflows/push-image.yml
pacostas: oops, I fixed it, the indentation was also wrong. I messed it up while committing it. Thanx for your time :) Feel free for any other suggestions, like the format of the registries.json file or if you think we should have that actions splitter into two one for the GCR and one for the dockerhub.

pacostas:
@ForestEckhardt @robdimsdale  I think this PR is ready to be landed.

ForestEckhardt (approved):

PR #1019, opened 2025-03-21
stack/get-usns general fixes/upgrades

pacostas:
## Summary

This PR does below things:
1. Adds noble on the regex expressions to properly scrap the usns from ubuntu page
2. Removes the default value from distribution, as it is better to error in case the user has not specified an ubuntu distribution compared to having a false fallback.
3. Upgrades `io/util` to `io`, as `io/util` is deprecated
4. Adds an extra flag called `packages_filepath` which is similar to packages and it enables passing the data through file. This is usefull in case there are a lot of packages, where in that case the workflow will fail due to terminal has a limit.

## Checklist

### DIFF
--- a/actions/stack/get-usns/entrypoint/main.go
+++ b/actions/stack/get-usns/entrypoint/main.go
@@ -44,12 +44,13 @@ type CVE struct {
 
 func main() {
 	var config struct {
-		Distro         string
-		LastUSNsJSON   string
-		Output         string
-		PackagesJSON   string
-		RSSURL         string
-		RetryTimeLimit string
+		Distro               string
+		LastUSNsJSON         string
+		Output               string
+		PackagesJSON         string
+		PackagesJSONFilepath string
+		RSSURL               string
+		RetryTimeLimit       string
 	}
 
 	flag.StringVar(&config.LastUSNsJSON,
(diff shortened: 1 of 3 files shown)


actions/stack/get-usns/entrypoint/main.go
```diff
@@ -76,6 +82,18 @@ func main() {
...
+	if distroExists == "" {
+		var validDistroValues = ""
+
+		for key := range distroToVersionRegex {
+			validDistroValues = validDistroValues + key + "\n"
+		}
+
+		errMessage := fmt.Sprintf("--distro flag has to be one of the following values: \n %s", validDistroValues)
+		log.Fatal(errMessage)
+	}
```
modulo11, line 95:
```suggestion
	_, ok := distroToVersionRegex[config.Distro]
	if !ok {
		log.Fatal(fmt.Sprintf("--distro flag has to be one of the following values: %v", slices.Sorted(maps.Keys(distroToVersionRegex))))
	}
```

Just a suggestion, these features are relatively new to the language and require a bump of the `go.mod`.

pacostas:
@modulo11 I upgraded the go version in order to support the new functions from go. Thank you for the review :)

modulo11 (approved):

PR #1148, opened 2025-08-15
Add multi arch support for implementation buildpacks

jericop:
## Summary

This PR updates implementation buildpacks to support multi-arch. All changes are backward compatible so existing single-architecture buildpacks are supported without any changes.

The first commit updates the scripts and workflows used to build, package, publish, and create draft releases.

The second commit adds a new reusable workflow for compiling dependencies and updates the `update-dependencies-from-metadata.yml` workflow to use it in favor of the composite action. The composite action does not support changing runners, which means compilation on non-native platforms has to be done using qemu emulation which can be very slow. The reusable workflow is currently setup to run on amd64 and arm64 runners to avoid emulation. If new platforms are added they will use emulation by default unless the workflow is updated to switch runners based on os and arch inputs. In order for a buildpacks to support multi-arch dependency the respective business logic in the dependency folder of each buildpack must be updated to add dependencies for different os and arch variants, as well as populating the os and arch metadata fields for each individual dependency. I will be updating cpython first and that can be replicated to other repos as needed.

The third commit changes the branch name that is looked for when creating PRs so that it matches the actual branch name where the commits to update buildpack.toml are pushed.

## Use Cases

We want multi-arch support for implementation buildpacks so we can eventually ship multi-arch composite buildpacks.

Testing was done with a fork of cpython and go-dist to verify buildpacks with and without compiled dependencies. Links below.

cpython single-arch-verify
https://github.com/jericop/cpython/actions/runs/16981769351
https://github.com/jericop/cpython/pull/9/files#diff-5f54cb49d0e28117084fad482c0daf89502d4ecb329230071fcf78e56dc939b3

cpython multi-arch-verify
https://github.com/jericop/cpython/actions/runs/16982432393
https://github.com/jericop/cpython/pull/10/files#diff-5f54cb49d0e28117084fad482c0daf89502d4ecb329230071fcf78e56dc939b3

go-dist single-arch-verify
https://github.com/jericop/go-dist/actions/runs/16984293635
https://github.com/jericop/go-dist/pull/11

## Checklist

### DIFF
--- a/implementation/.github/workflows/compile-dependency.yml
+++ b/implementation/.github/workflows/compile-dependency.yml
@@ -1,0 +1,106 @@
+name: 'Compile Dependency on Target - Reusable Workflow'
+
+description: |
+  Compiles Dependency on given target, os, and arch
+
+on:
+  workflow_call:
+    inputs:
+      version:
+        description: 'dependency version'
+        required: true
+        type: string
+      target:
+        description: 'dependency OS target variant'
+        required: true
+        type: string
+      os:
+        description: 'platform OS (e.g., linux)'
+        required: true
+        type: string
+      arch:
+        description: 'platform architecture (e.g., amd64)'
+        required: true
+        type: string
+      shouldCompile:
+        description: 'whether to compile the dependency'
+        required: true
+        type: boolean
+      shouldTest:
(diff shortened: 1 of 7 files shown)


pacostas (approved):
LGTM

PR #1209, opened 2025-10-30
Add renovate config for buildpacks

nicolasbender:
## Summary

This serves as a basic configuration for dependency updates via renovate for buildpack repositories. Currently, this replaces dependabot updates only, so every dependency managed by paketo-bot needs some more investigation.

## Checklist

### DIFF
--- a/renovate/renovate-buildpack.json
+++ b/renovate/renovate-buildpack.json
@@ -2,19 +2,5 @@
   "$schema": "https://docs.renovatebot.com/renovate-schema.json",
   "extends": [
     "github>paketo-buildpacks/github-config//renovate/renovate-default"
-  ],
-  "customManagers": [
-    {
-      "customType": "regex",
-      "fileMatch": [
-        "^buildpack.toml$"
-      ],
-      "matchStrings": [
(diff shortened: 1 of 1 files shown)


pacostas:
Also I think that upgrading the API version of the buildpack.toml to its latest, might not be very useful, as for each buildpack there is migration that needs to be done, between each api version. I would suggest if is feasible, instead of always going to the latest version, to go to the next version from what currently the api is.

pacostas:
I agree having a repo that stores all the api versions. Do you think we can also fetch the api versions from the release https://github.com/buildpacks/spec/releases with the gh cli? The release notes seem pretty consistent and updating the content of the `paketo-buildpacks/cnb-buildpack-api-version` might need another github action that probably will fetch the content from the releases or it might have to be maintained manual.

In terms of having all buildpacks at the same api version, I think this might be blocking for other buildpacks from upgrading to a latest version or a much grater than what renovate might suggest. E.g. for node-engine, where we use the api 0.10 compared to the majority of the buildpacks that are on version 0.7. I think it would be preferable if each buildpack could upgrade separately, without having to follow the same api as the rest of the buildpacks.

PR #1266, opened 2026-01-19
Usns issue fix

pacostas:
## Summary

This PR:
* Instead of fetching the usns from the feed url, it fetches them from the ubunut api https://ubuntu.com/security/notices.json
* I separate the action.yml with a bash script on the generate package list action, for easier development.

This change is necessary as the feed url, shows only the last 10 usns. This is a problem due to the workflows run once every day, as a result, the workflows might loose some of the changes due to the last day might have occured more than 10 usns. By switching to ubuntu api, except the fact that is faster due to we make much less http requests, we have the ability to fetch any number of the latest usns. Specifically, we fetch the last 40 usns and the usns that include the distro and not for any distro.  

## Checklist

### DIFF
--- a/stack/.github/workflows/create-release.yml
+++ b/stack/.github/workflows/create-release.yml
@@ -126,7 +126,7 @@ jobs:
             support_usns=$( jq '.support_usns' ${{ env.STACKS_FILEPATH }} )
           fi
 
-          if [ $support_usns == true ]; then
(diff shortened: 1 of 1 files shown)


pacostas:
Thank you for reviewing it, I'll do some more testing and if everything runs ok, I'll land it.

